PR #59, opened 2015-06-10
Add an option to emit warning if function length is greater than a threshold

clementleger:
In order to enforce good programming practice, add the possibility to emit a warning if function length is greater than a certain size. Also display the function length to ease reading of results.

### DIFF
--- a/lizard.py
+++ b/lizard.py
@@ -33,6 +33,7 @@ if sys.version[0] == '2':
 VERSION = "1.8.11"
 
 DEFAULT_CCN_THRESHOLD = 15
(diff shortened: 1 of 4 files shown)


terryyin:
Hi @clementleger ,
Sorry for keeping this so long. I'm merging year code now.
May I ask what is the purpose of valid/hudson/build.rb?
And I will probably change the default to 1000, not to surprise the current users.

terry

clementleger:
Hi Terry,
I made a mistake by commiting the build.rb file.
I will do another cleaner request :)

PR #63, opened 2015-10-05
Processing with C++11 'noexcept' keyword

rakhimov:
This fix mainly applies to constructors with initialization lists.
The last member variable name is picked up as the name for reporting.


terryyin:
thanks rakhimov!

Issue #64, opened 2015-10-05
C++11 Uniform Init. within constructor Initialization list


rakhimov:
C++11 '{}' for uniform initialization
in constructor initialization list
seems to break some assumptions in the code
that the definition starts with the first brace.

terryyin:
Hi,
I really need to learn C++11 more systematically:-)
It's solved and I'm going to create a new release.

Terry

rakhimov:
Thanks for fixing it fast.

PR #65, opened 2015-10-05
Add tests for failing constructor C++11 init list

rakhimov:
Tests are examples for issue #64.
If a member variable or base class is initialized
with an initializer list or uniform initialization
within the initialization list,
the constructor is parsed as multiple function definition.


terryyin:
I've merged your tests and fixed them.

This time with a bit different approach. Instead of adding another _state_, I added a OneInitializerState class as  a 'Command Pattern'. I think I'm going to gradually refactor the code to use State classes instead of _state_ method. It's easier to remove duplicated functionalities, have nested structure, and keep a local state.

Issue #11, opened 2013-09-09
name of destructors are not correct


terryyin:
They comes without ~

rakhimov:
@terryyin , isn't this already fixed?

PR #67, opened 2015-10-06
Hack '~' into C++ destructor names

rakhimov:
This is a quick and dirty fix.
@terryyin may hack further to find the right way (state, regex?).


rakhimov:
This should address issue #11

terryyin:
Merged and replaced the hack with a "proper" solution.

Thanks for all the small fixings.

PR #68, opened 2015-10-06
C++11 '>>>' closing brackets problem

rakhimov:
In function declaration,
deeply nested templates may end with multiple closing bracket
without spaces in-between (C++11)
```cpp
 void f(a<b<c<d>>> arg) {}
```
This is conflicting with '>>>' symbol.
Whatever that symbols means,
its deletion seems not to cause any test failures.


terryyin:
Oops! So templated!

rakhimov:
It turns out Nvidia Cuda C/C++ code uses this ``<<<`` along with ``>>>``` for kernel launch.
It sounds awkward, but probably somethings that makes sense and is a syntactic sugar for GPGPU developers.

From Wikipedia:
```cpp
kernel <<< gridDim, blockDim, 0 >>> (d_data, height, width);
```
I am not sure if Lizard should support the code.

Got the following quote from [GPGPU language comparison blog](https://streamcomputing.eu/blog/2016-04-05/comparing-syntax-cuda-opencl-hip/)

> CUDA’s “<<< >>>” breaks all C/C++ compilers, making it very hard to make a frontend of IDE-plugin.

terryyin:
I guess the cost is it will slow the tokenising a little bit.

Issue #70, opened 2015-10-16
Qualify C++ functions with namespace and class names


rakhimov:
Currently, the following code will produce the same function signatures for reporting.
```cpp
struct A { bool operator()(int) {...} };
struct B { bool operator()(int) {...} };
```
It would be beneficial to have class name and/or namespace attached to the function names.

rakhimov:
It is not clear if all instances of function definitions should be qualified. There are some tricky cases with C++.

- Class/struct within a function (the qualification may need to include the function name):
```cpp
void foo() { struct A { void bar() {} }; }
```
- Lambda functions (unnamed closure/class in C++11):
```cpp
void foo() { auto bar = [](){ return 42; } }
```

terryyin:
@rakhimov , this looks like JavaScript!
Hmmm, I've actually been thinking what measurement would make sense for JavaScript nested functions, which aren't corner cases at all for js.
So far, lizard will just ignore them. But I think maybe we should add the ability of detecting them before we can decide how to measure.

rakhimov:
Can the complexity of the inner constructs be just silently added to the complexity of the outermost construct?

rakhimov:
The Lizard analysis of Python code with nested functions reports each function separately with line number information. Is this the same behavior across other languages?

terryyin:
Adding to the outermost construct makes a lot of sense. Let me do it.
The Python nested function was implemented separately, unfortunately. I will see if I can merge the implementation.

tobias-klein:
@terryyin Has there been any work on this since last October? I think this would be a very useful enhancement.

terryyin:
Hi @tobias-klein thanks for getting back to this. The last Oct 17 commit already does most of the namespace work https://github.com/terryyin/lizard/commit/eb95a74ae2ac02763a7f4f6736f34225cbca69a6 .

And now it is like @rakhimov said that the inner structure or lambda add complexity to the outermost function.

But if you are referring to the javascript nesting, there hasn't been any progress yet. So what do you want?

rakhimov:
I feel sorry for C++ compiler designers.
Working on 'nesting_level' and 'nested_structures' metrics,
I have been discovering horror-inducing constructs that is possible in C++.
These constructs relate to this issue as well.
Consider this example:

```cpp
  for (struct {int s;int foo() { while(true) break;  return 42;}} a{42};;) break;
```

That's unnamed struct that can have arbitrary complexity
in addition of being inside for-loop declaration.
[...]

PR #71, opened 2015-10-16
Add a test for inline C++ operator overload

rakhimov:
The name of the class is not attached to the operator overload.
This is an example for #70

[Review on Reviewable](https://reviewable.io/reviews/terryyin/lizard/71)


terryyin:
Hi @rakhimov , I've passed your test and added a few more. I did it for namespaces too.

What do you think about adding namespaces to the function names? useful or too noise?

rakhimov:
I think it is a good idea for C++ code. For large codebase with multiple namespaces, it is helpful for reports to get the namespace attached to function names, i.e., ```namespace::class::function```.
This can be optional for small codebase.
I would suggest Lizard have a flag for this optional feature.

PR #73, opened 2015-10-19
C++ namespace alias failure

rakhimov:
Relates to issue #70

[Review on Reviewable](https://reviewable.io/reviews/terryyin/lizard/73)


terryyin:
@rakhimov thanks! this is actually a bug. I fixed it with a hack. Will refactor it to a proper solution next.

PR #75, opened 2015-10-22
Ignore C++11 unnamed namespace for qualifications

rakhimov:
Nesting unnamed namespace adds extra ``::`` to the qualification.
```cpp
namespace real { namespace { void foo(int) {...} } }
```
The function is qualified as ``real::::foo`` instead of just ``real::foo``.

[Review on Reviewable](https://reviewable.io/reviews/terryyin/lizard/75)


terryyin:
Thanks! I wish Python has the Ruby list.compact method:-)

PR #79, opened 2016-01-08
Exclude code in C/C++ 'assert' from CCN

rakhimov:
Consider a simple use case:

```cpp

assert((conditional_expression) && "More description...");

```

This line adds 1 + CCN(conditional_expression)
if the assertion is counted as a part of a function.

Since assertions are mostly used
to check (pre, post)-conditions,
to explicitly state assumptions,
and to catch programmer errors,
they are not expected to be a part of the flow of a function.
Moreover, assertions with side-effects is a very bad idea.
The function should stay the same
if all the assertions are removed (in release mode, for example).

[Review on Reviewable](https://reviewable.io/reviews/terryyin/lizard/79)

### DIFF
--- a/lizard.py
+++ b/lizard.py
@@ -241,6 +241,7 @@ class FileInfoBuilder(object):
         self.newline = True
         self.global_pseudo_function = FunctionInfo('*global*', filename, 0)
         self.current_function = self.global_pseudo_function
+        self.ignore_complexity = False
 
     def add_nloc(self, count):
         self.current_function.nloc += count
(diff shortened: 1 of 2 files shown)


rakhimov:
The implementation currently breaks the decorator based ```_state_imp``` function design.
@terryyin , any hints how to integrate this PR gracefully?

terryyin:
@rakhimov thanks for the pull request. I get your idea:-)

Yeah, I've been refactoring the state-machine to make it more modularised. The new design isn't quite completed yet.

I'm thinking to implement it with an "extension" first for now...

Any way. Let me merge and then show you my idea.

terryyin:
I created a new extension (lizard_ext/lizardignoreassert.py), which can be used with:

    lizard -EIgnoreAssert

https://github.com/terryyin/lizard/commit/5d23f3e08470d76703fe846357566971c357bdd0

rakhimov:
I like this approach. Thanks for putting extra work to make this PR fit.

PR #80, opened 2016-01-10
Add test cases for IgnoreAssert extension

rakhimov:
This is additional tests for handling C/C++ asserts with an extension #79.
The code should ignore complexity of any code inside assert.

[Review on Reviewable](https://reviewable.io/reviews/terryyin/lizard/80)

### DIFF
--- a/test/testAssertionExtension.py
+++ b/test/testAssertionExtension.py
@@ -11,3 +11,16 @@ class TestAssersionExtension(unittest.TestCase):
         result = get_cpp_function_list_with_extnesion(
                 "void fun() { static_assert(a && b && c, \"Failed\"); }", LizardExtension())
         self.assertEqual(1, result[0].cyclomatic_complexity)
+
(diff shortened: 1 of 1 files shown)


terryyin:
I've passed the test and created a new release.

rakhimov:
I think the current implementation is practical enough but not general.
It is still failing to ignore *any* code complexity inside of assert. It is weird but possible to put "while, for, if, goto, switch, function, class, etc." through lambda.

terryyin:
Hmm, I think if extremely complicated expression is used in assertion, probably it shouldn't be ignored:-)

rakhimov:
I think consistency is better. It is hard to come up with and justify qualifications for the code inside of asserts, but with the extension approach, users can decide when not to ignore asserts.

terryyin:
Good point.

And it should work with C/C++ code. Let me do it properly.

PR #93, opened 2016-02-09
Use Python3 division for calculation of averages

rakhimov:
In some places, it seems that the results of calculations
are expected to be floats,
but integer division of Python2 is used.
In other places, explicit cast to float is made to get the float result.
Python3 division approach for integers helps cleanup these places
and make them consistent.

If Lizard relies on integer division approach of Python2,
which doesn't seem to be the case,
the code must be corrected with ``//``.

[Review on Reviewable](https://reviewable.io/reviews/terryyin/lizard/93)


terryyin:
hi @rakhimov ,
Wow, the code becomes much more clean now. Thanks.

PR #100, opened 2016-02-29
Nesting Depth

mehrdad89:
Introducing Maximum Nesting Depth for Clike and Python languages, which detects the maximum 'nestedness' (nesting for-loops, if-statements, etc.) in any source code without caring about the C/C++ header files or Python imports.

### DIFF
--- a/.idea/dictionaries/emehmeh.xml
+++ b/.idea/dictionaries/emehmeh.xml
@@ -1,0 +1,8 @@
+<component name="ProjectDictionaryState">
+  <dictionary name="emehmeh">
+    <words>
+      <w>nindent</w>
+      <w>patricksmacchia</w>
+    </words>
+  </dictionary>
+</component>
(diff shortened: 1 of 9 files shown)


mehrdad89:
I will try to fixed them as soon as possible. Sorry about that!

terryyin:
Please take time:-)

mehrdad89:
Thanks, I will fix the issues then I will send a pull request again.

terryyin:
Hi @mehrdad89 , just to let you know that during the refactoring I had to revert your changes in the lizard_languages/python.py, because I found yielding a <indent> might not be a best way of handling it. And it breaks other feature as well.

I will bring it back probably with another solution after the refactoring.

mehrdad89:
@terryyin Great, I'll be waiting eagerly to see what you can up with. I also faced another trouble regarding adding structural fan-in & fan-out. I am now looking to see how I can structural fan-in and fan-out at a local scope while having global scope in mind.

Issue #98, opened 2016-02-26
cyclomatic complexity (&&) bug in clike


mehrdad89:
HI! 

I am currently writing tests to find bugs in nesting depth metric which I soon commit to this project however as of now (lizard 1.10.4) there is a problem with cyclomatic complexity counting down (&&) in this unit test.

    def test_one_function_with_statement_no_curly_brackets(self):
        result = get_cpp_function_list("""
        x c() {
        if (a > -1 && b>= 0 )
              if(a != 0)
                    a = b;
        }
        """)
        self.assertEqual(4, result[0].cyclomatic_complexity)

I think it has a direct contradiction with the (test_one_function_with_r_value_ref_in_body) unit test .

terryyin:
Hi @mehrdad89 sorry I overlook your message (the preview of the email has only "Hi!") I'm looking at it...

terryyin:
@mehrdad89 you are right, the bug is introduced when I made the solution for r_value_ref. It's now fixed with https://github.com/terryyin/lizard/commit/2a165bdd1cb1b318a15bf6b1bfdce8bcc32d70df
I will create a new release next week.

thanks for the reporting.

mehrdad89:
thank for the reply! 
Actually I just finished coding a new metric (nesting depth metric) into lizard.
At the moment, it is only supporting clike and python. Hopefully if all test are done I will be pushing it to your repository today.

terryyin:
Great! Cannot wait (but it’s bed time here).

BTW, adding tests after the implementation sounds really boring:P

mehrdad89:
Yeap! tell me about it!

Issue #69, opened 2015-10-06
Detection of Deeply Nested Control Structures


rakhimov:
This metric may not apply to the whole function,
but the maximum 'nestedness' (nesting for-loops, if-statements, etc.)
may be an interesting metric to detect code smell.

It closely relates to indentation. 
Got this from the Linux kernel coding style:
>The answer to that is that if you need
more than 3 levels of indentation, you're screwed anyway, and should fix
your program.

mehrdad89:
Hi @terryyin! 
I think this metric will be a great addition to this awesome tool. I'm gonna do it!
Hopefully, I can finish it this weekend.

PR #101, opened 2016-03-01
Nesting Depth

mehrdad89:
Introducing Maximum Nesting Depth for Clike and Python languages, which detects the maximum 'nestedness' (nesting for-loops, if-statements, etc.) in any source code without caring about the C/C++ header files or Python imports.


rakhimov:
@mehrdad89, thanks for putting work on #69.
I am interested in this metric and test running your implementation.
I am curious why the default nesting depth number is 7.
It seems too high. I think 3 would be more reasonable.

test/testNestingDepth.py
```diff
@@ -0,0 +1,119 @@
...
+
+    def test_one_function_with_one_nd_condition(self):
+        result = get_cpp_function_list("int fun(){if(a){xx;}}")
+        self.assertEqual(1, result[0].max_nesting_depth)
+
+    def test_one_function_nd_with_question_mark(self):
+        result = get_cpp_function_list("int fun(){return (a)?b:c;}")
+        self.assertEqual(1, result[0].max_nesting_depth)
+
+    def test_one_function_nd_with_forever_loop(self):
```
rakhimov: Is this a special case?

test/testNestingDepth.py
```diff
@@ -0,0 +1,119 @@
...
+                }''')
+        self.assertEqual(1, len(result))
+        self.assertEqual(2, result[0].max_nesting_depth)
+
+    def test_one_function_nd_with_r_value_ref_in_parameter(self):
+        result = get_cpp_function_list("int make(Args&&... args){}")
+        self.assertEqual(0, result[0].max_nesting_depth)
+
+    def test_one_function_nd_with_r_value_ref_in_body(self):
+        result = get_cpp_function_list("int f() {Args&& a=b;}")
```
rakhimov: Is ```&&``` r-val ref tested with the bug #98 in it?
(the ND must be 1 instead of 0)

mehrdad89:
No reason, I thought it would be a little too demanding especially if you are writing a complex function, but I think you're right. I will change it to 4 or maybe 5. and yes I will fix the r-Val ref bug I wrote that one before @terryyin fixed that issue.

test/testNestingDepth.py
mehrdad89: Question mark is interpreted as an if statement in [https://github.com/terryyin/lizard/pull/101/files#diff-7e6e46eb7ddafa74e71930de5e50f4ceR15](url)

rakhimov: I mean ```test_one_function_nd_with_forever_loop```.

mehrdad89: Oh, that one... yes it is considered as one forever loop just like what @terryyin has considered in cyclomatic tests cases.

rakhimov: I am in doubt if this should be special case for ND.
For CCN, it makes sense because the control path will be executed no matter what; there is no branching.
It is like ```if (true)```.

mehrdad89: I think it can be considered same as an infinite while loop `while( true )`. From what I understood, I thought an infinite loop should still be considered as one ND. Well, I have to read some text to confirm your statement. Do you have any references talking specifically about this issue?

rakhimov: I am not sure if there has been a definitive research in this area,
but I found this http://www.ndepend.com/docs/code-metrics#ILNestingDepth.
It seems to be a suite for Visual Studio.
Their interpretation of the ND seems to be more like local maximum CCN.
For example,
```cpp
if (a && b) {}
```
the statement has ND 2.
If this is the definition of the ND you are implementing,
your test as written looks correct to me.
However, I think this definition of ND doesn't add much value to the existing facilities that calculate CCN.
I suspect that this ND metric has a high correlation with the CCN,
so it seems redundant for me.

The issue #69 is about nesting control structures,
which closely relates to indentation,
so if you need to indent your code deeper than 3 tabs, structures, (i, j, k loops),
your code has problems.
I think this metric or smell detector is more useful in practice.
It is simple and easy to internalize for developers to act upon or refactor the code.

mehrdad89: I will see if I can look deeper into this issue but I think that calculation of branching in a function is called "branching complexity sneed metric". If you wanna talk about correlation, you can find more similarities between metrics such as Halstead complexity to CCN than nesting depth to CCN.

lizard.py
```diff
@@ -41,11 +42,10 @@
 except ImportError:
     pass
 
-
 VERSION = "1.10.4"
 
-DEFAULT_CCN_THRESHOLD, DEFAULT_WHITELIST, \
-    DEFAULT_MAX_FUNC_LENGTH = 15, "whitelizard.txt", 1000
+DEFAULT_CCN_THRESHOLD, DEFAULT_ND_THRESHOLD, DEFAULT_WHITELIST, \
+    DEFAULT_MAX_FUNC_LENGTH = 15, 7, "whitelizard.txt", 1000
```
terryyin: @mehrdad89 instead of changing directly in lizard.py and clike.py, I'd like these changes to be in an extension first (see lizard_ext/lizardexitcount.py for example). Then use command like:

    lizard -E nd

To include the new metrics. We can have the metrics in the default result collection when it's stabilised later.
You can make the change, or I can accept your pull request and make the change for you. What do you prefer?

test/testNestingDepth.py
terryyin: Hmm, @rakhimov 's comment makes me wonder if forever loop should be counted as one CCN...

    while (i < 10) {}

is the same as:

    for(;;) {if (i < 10) break;}

So

    for(;;){}

shouldn't be counted probably. What do you think?

lizard.py
mehrdad89: If you don't mind, make the changes you deem appropriate. I'll be happy.

test/testNestingDepth.py
mehrdad89: I'm thinking what he's saying is true but CCN and Nesting Depth metrics never mentioned infinite loop. And your example reveals the problem in our algorithm for these metric.
We should fix this issue and modify our code. We can ignore infinite loop statement where they have hard coded value of `true` in them. But we still can not identify some of them in some cases, not just `for(;;){}`. 

For Example:
`   a=1
    if (a==1){
        do_something()
    }`
Can you think of a solution other than preventing some of those and catch specific cases we can identify in every language?

terryyin: I've opened another issue for this https://github.com/terryyin/lizard/issues/103

Issue #92, opened 2016-02-04
suport PHP


terryyin:

iconnor:
@terryyin this would be great!

mikesive:
@terryyin seconded

tobias-klein:
@terryyin Yes, please add that! :)

terryyin:
C# has been added. I'm going to close this issue and see if anybody would create an issue for the missing of PHP.

mikesive:
Hey @terryyin I am really keen for PHP support as PHPMD is very slow. Is it possible to do PHP as well? I had hoped the bounty on bountysource applied to both PHP and C#.

terryyin:
OK, I've reopened the issue and change the title.

@mikesive I'm going to give PHP a try, but can not be very sure this time. Because PHP can be also inline in HTML. All the other languages are tokenised as the fist step. But I guess with PHP probably some other preprocessing need to be done before tokenising.

we'll see:-)

mikesive:
Thanks @terryyin look forward to seeing how you go.

terryyin:
@mikesive , I've added PHP and created a new release in PyPI (1.12.0).
It will be great if you can help to verify it.
If you want to run lizard for PHP only, do

    lizard -lphp

I will improve the PHP support to add things like "namespace" (class names). But I think the basic function is there already now.

The commits did the job are:
https://github.com/terryyin/lizard/commit/69336258c2f1cf4c3592a1fa44da6d0091459865
https://github.com/terryyin/lizard/commit/539566d37862902c4068e14e0a7d732827469a99
https://github.com/terryyin/lizard/commit/0ae15e11301a9a79cab2d89bc13e2b0c2bf804a7

mikesive:
@terryyin looks good! I had a quick run through today, very fast compared to other PHP tools. I will test more thoroughly tomorrow. Thanks for getting this done!

iconnor:
This works for us! I think you can close it now @terryyin

terryyin:
Great!

PR #104, opened 2016-03-09
Add C# test file #92

iconnor:
I know it is probably not that much help, but I hoped it would show we are keen for this feature. I took the example code from: https://msdn.microsoft.com/en-us/library/ms182212.aspx?cs-save-lang=1&cs-lang=csharp#code-snippet-1

### DIFF
--- a/test/testCsharp.py
+++ b/test/testCsharp.py
@@ -1,0 +1,81 @@
+import unittest
+from lizard import analyze_file, FileAnalyzer, get_extensions
+
+
+def get_csharpe_fileinfo(source_code):
+    return analyze_file.analyze_source_code("a.cs", source_code)
+
+
+def get_csharpe_function_list(source_code):
+    return get_csharpe_fileinfo(source_code).function_list
+
+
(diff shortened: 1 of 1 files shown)


terryyin:
Hi @iconnor thank you very much, I think this is very useful. I will be working on it soon, I think.

terryyin:
Hi @iconnor , passing those tests turned out to be extremely easy, as C# syntax is very similar to C++. Now I'm looking into detailed C# syntax and see if there's any exception (I'm a lousy C# programmer myself).

I have a question regarding the CCN for C#. Should the Null Coalescing Operator (??) be counted as one condition? I'm guessing yes.

iconnor:
As the alternative is an if null statement, I would vote yes on ?? increasing complexity. It is still something you have to think about when reading the method. Doing it on one line is neat but still a condition.

terryyin:
Hi @iconnor ,
I've made a new release (1.11.0) with support for C#. Could you please test it a bit?

PR #108, opened 2016-03-17
Introducing fan_in _fan_out prototype 

mehrdad89:
Structural fan in fan out in Lizard. (logic will be improved)


mehrdad89:
@terryyin suggested that structural fan in and fan out should be moved into another file named `lizardfans.py`. Tests will be added pretty soon afterwards.

PR #110, opened 2016-03-23
Structural fan_in _fan_out modifications

mehrdad89:
@terryyin I think I fixed the given issues based on your feedback.
In the case of `stderr`, could you take a look it? hopefully, I understood what you meant by that comment.

### DIFF
--- a/lizard.py
+++ b/lizard.py
@@ -470,18 +470,17 @@ class OutputScheme(object):
         self.items.append({'caption': " location  ", 'value': 'location'})
 
     @staticmethod
-    def get_stringtype():
+    def is_string_instance(ext):
         try:
             stringtype = basestring
         except NameError:   # Not compatible with python 3
             stringtype = str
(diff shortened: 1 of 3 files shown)


lizard.py
```diff
@@ -469,13 +469,18 @@ def __init__(self, ext):
...
+
     def _ext_member_info(self):
         for ext in self.extensions:
             if hasattr(ext, "FUNCTION_CAPTION"):
-                try:
-                    stringtype = basestring
-                except NameError:   # Not compatible with python 3
-                    stringtype = str
+                stringtype = OutputScheme.get_stringtype()
                 if isinstance(ext.FUNCTION_INFO_PART, stringtype):
```
terryyin: instead of get_stringtype, we can probably extract a is_string_instance(). What do you think?

mehrdad89: yes , that would be much better.

PR #112, opened 2016-04-07
improving lizardfans.py functionality & adding tests support for Windows platform

mehrdad89:
adding support for overloading functions and reducing the execution time on lizardfans.py extension for
lizard_ext/lizardfans.py

    Total nloc   Avg.NLOC  AvgCCN  Avg.token   Fun Cnt  Warning cnt   Fun Rt   nloc Rt
    ------------------------------------------------------------------------------------------
        38792      23.8     4.7      138.8     1407           67      0.05    0.24
    --- 14.48589921 seconds ---

to:  

    Total nloc   Avg.NLOC  AvgCCN  Avg.token   Fun Cnt  Warning cnt   Fun Rt   nloc Rt
    ------------------------------------------------------------------------------------------
      38792      23.8     4.7      138.8     1407           67      0.05    0.24
    --- 4.04137706757 seconds ---

I also added support for tests in Windows platform.
The next step is to add the external scope for Structural fan-out.


terryyin:
@mehrdad89 have you tried with C++ functions that comes with namespace? E.g. MyClass::foo.

I'm thinking perhaps I should remove the namespace from "name" and put it only in the "long_name"

lizard_ext/lizardfans.py
```diff
@@ -27,28 +32,35 @@ def __call__(tokens, reader):
...
         """
         Preparation of calculating the fan in and fan out (prototype)
         """
+        infile = Infile()
         self.name_list = [func.name for func in fileinfo.function_list]
         for func in fileinfo.function_list:
-            body = self.method_open(func)
+            if not func.filename == infile.name:
+                infile = self.open_method(func.filename)
+            body = self.store_method(infile, func)
```
terryyin: The name "store_method" is wrong. It didn't store anything. It returns the method body, so probably "get_method_body" or just "method_body" would be a better name?

lizard_ext/lizardfans.py
```diff
@@ -10,6 +10,11 @@
 
 import re
 from sys import stderr
+import itertools
+
+
+class Infile(object):  # pylint: disable=R0903
```
terryyin: This is a "lazy class", you don't need it :-)

lizard_ext/lizardfans.py
```diff
@@ -27,28 +32,35 @@ def __call__(tokens, reader):
...
         Preparation of calculating the fan in and fan out (prototype)
         """
+        infile = Infile()
         self.name_list = [func.name for func in fileinfo.function_list]
         for func in fileinfo.function_list:
-            body = self.method_open(func)
+            if not func.filename == infile.name:
+                infile = self.open_method(func.filename)
+            body = self.store_method(infile, func)
             self.calculate_fan_in_fan_out(self.name_list, body, fileinfo, func)
```
terryyin: The name "calculate_fan_in_fan_out" is "lying." It doesn't just calculate, it also updates the state. So, I'm suggesting either:
1. rename it, or
2. Split the calculating and updating into two methods, to be come something like: self.update_fans(self.calcuate_fans(...), ...)

If you do the 2nd one, perhaps you will see a new potential class emerges.

test/testFans.py
```diff
@@ -56,8 +56,28 @@ def test_fans_initiation(self):
     def test_empty_body(self):
         result = get_cpp_function_list_with_fans_extension("a.cpp",
                                                            "int fun(a){})")
-        body = self.ext.method_open(result[0])
-        self.assertEqual([], body)
+        body = self.ext.open_method(result[0].filename)
+        self.assertEqual(None, body)
+
+    def test_open_function_infile(self):
```
terryyin: Please avoid reading and writing files in test if possible. What about mock the "open" method?

lizard_ext/lizardfans.py
mehrdad89: you're right, get_method_body would be more suitable.

test/testFans.py
mehrdad89: ok, I will change them soon then send a new pull request soon.

lizard_ext/lizardfans.py
mehrdad89: Of course, Infile class will be removed.

lizard_ext/lizardfans.py
```diff
@@ -57,9 +69,7 @@ def calculate_fan_in_fan_out(name_list, body, fileinfo, func):
             for i, name in enumerate(name_list):
                 re.split(r'; |, |\*|\n', line)
```
mehrdad89: I need to get acess to your codereader to get the tokens and replace this line of code. Can I call a function to it so it will get rid of the useless information nad make the code cleaner?

terryyin: Sure, let's first make it work. Then make it right.

lizard_ext/lizardfans.py
terryyin: You may try to make self.open_method(func.filename) into the __init__ of the InFile class and move the static methods to the InFile class.

mehrdad89: I just found a way to remove the infile class and move most of them into outside of the class. It is much simpler this way. I just have to fix calculate_fan_in_fan_out method and the tests to get it done. 
Thank for the suggestion by the way.

PR #119, opened 2016-05-05
Implement Nested Structure counting for C++

rakhimov:
The description of the metric is in Issue #69 and in the documentation.

---
This change is [Reviewable](https://reviewable.io/reviews/terryyin/lizard/119)

### DIFF
--- a/lizard_ext/lizardns.py
+++ b/lizard_ext/lizardns.py
@@ -7,10 +7,11 @@ originally written by Mehrdad Meh and Terry Yin.
 """
 
 from lizard import FileInfoBuilder, FunctionInfo
-from lizardnd import patch, patch_append_method
+from lizard_ext.lizardnd import patch, patch_append_method
 
 
 DEFAULT_NS_THRESHOLD = 3
+
 
 class LizardExtension(object):  # pylint: disable=R0903
 
@@ -67,14 +68,16 @@ class LizardExtension(object):  # pylint: disable=R0903
 
                 elif token == structure_end:
                     reader.context.pop_brace()
-                    reader.context.pop_nested_structure();
+                    reader.context.pop_nested_structure()
 
                 elif token == indent_indicator:
-                    reader.context.pop_nested_structure();
+                    reader.context.pop_nested_structure()
 
             yield token
 
 
+# TODO: Some weird false positive from pylint. # pylint: disable=fixme
+# pylint: disable=E1101
 class NSFileInfoAddition(FileInfoBuilder):
 
     def add_nested_structure(self, token):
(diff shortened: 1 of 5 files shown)


mehrdad89:
Great addition! let's see what @terryyin think about the changes!

test/testNestedStructures.py
```diff
@@ -0,0 +1,211 @@
...
+        result = process_cpp("""
+        x c() {
+          {{{{{{for (a : c) {
+            {{{{if(a != 0) {{{{
+                a = b;
+            }}}}}}}}
+          }}}}}}}
+        }
+        """)
+        self.assertEqual(2, result[0].max_nested_structures)
```
terryyin: Why the nesting with 'bare brackets' are ignored? There usually need to be good reason for having them. And in C++, they actually trigger destructors.

terryyin:
great! @rakhimov , I like it very much.

So far this implementation works only for "C-like" languages. I have implemented some more general data for nesting.
At any time, you can get the current nesting level via `context. self.current_nesting_level` and
`context.current_function.top_nesting_level` keeps the top nesting level of the current function. Perhaps you can refactor the code to use these two values to calculate the NS result. Then it will become a general implementation for all supported languages.

I like the theory very much. I or somebody else should add CCN to it.

lizard_ext/lizardns.py
```diff
@@ -0,0 +1,137 @@
...
+        """
+        structures = set(['if', 'else', 'foreach', 'for', 'while', 'do',
+                          'try', 'catch', 'switch'])
+
+        structure_indicator = "{"
+        structure_end = "}"
+        indent_indicator = ";"
+
+        for token in tokens:
+            if reader.context.is_within_structure():
```
terryyin: If you do everything in this one method, there's no need to add information to `reader.context`. A local variable/object in this method might serve the purpose already, if the state is not going to be shared with other processors.

rakhimov: That indeed simplifies the code.
By the way, is it possible to get brace/bracket counts from the reader?

rakhimov:
Hi @terryyin ,
you have merged this PR too early.
It is quiet annoying to submit patches to patches.
I would commit additional patches within this PR to get your reviews into code,
which makes it more coherent.
I wish GitHub had 'unmerge' feature.

test/testNestedStructures.py
rakhimov: Technically it is not control-flow structure,
but I think it is a good idea to include bare scope into this metric.
However, the bare braces have more to it than control-flow structures.
For example, wrapping your code with openmp pragma to get it parallelized when built with openmp; the code that does the job stays the same for all builds.

rakhimov:
``context.current_nesting_level`` and ``current_function.top_nesting_level`` are undocumented.
It is not clear what ``nesting level`` means.
For example, the ``current_nesting_level`` produce different nesting for the following code:

```cpp
if (x) while(y) continue;

if (x) { while(y) { continue } }
```
In the first line of code, all ``if``, ``while``, ``continue`` get nesting 1.
In the second line of code, the nesting output is: (``if``, 1), (``while``, 2), (``continue``, 3).
Is it just brace counting for C++?

PR #120, opened 2016-05-08
Generalize Nested Structure Counting for languages 

rakhimov:
The code is refactored to work with nesting level metric
provided by language readers of Lizard.
The feature is only tested with Python and C++.
This implementation should be enough to close #69.

PR #123, opened 2016-05-13
Implement C++ nesting_level for structures w/o braces

rakhimov:
The 'nesting_level' is considered for internal uses only
and necessary for properly implementing
generic Nested Structures metric (PR #120).
The implementation is tested indirectly through the aforementioned metric.

---
This change is [Reviewable](https://reviewable.io/reviews/terryyin/lizard/123)

### DIFF
--- a/lizard_languages/clike.py
+++ b/lizard_languages/clike.py
@@ -72,16 +72,85 @@ class CppRValueRefStates(CodeStateMachine):
         self.context.add_condition(-tokens.count("&&"))
 
 
+# pylint: disable=R0903
 class CLikeNestingStackStates(CodeStateMachine):
-    # pylint: disable=R0903
+    """Machinery to track nesting levels of tokens.
+
+    The main indicators of nesting are braces: '{' and '}'.
+    However, the code is complicated due to braceless structures,
+    which are control-flow structures in C-like languages.
+    Moreover, using braces as nesting level indicators leads to a big caveat;
+    C++ initializer lists, uniform initialization,
+    and any other constructs with braces add extra nesting level.
+
+    Another complication comes from nested classes inside function bodies
+    and control-flow declarations/bodies.
+    The handling of these complex cases is unspecified and can be ignored.
+    """
+
+    # Beasts that can be defined within one line without braces.
+    __braceless_structures = set(['if', 'else', 'for', 'while', 'do',
+                                  'switch'])
+    __paren_count = 0  # Used only to tackle the beasts.
+    __braceless = None  # Applies only to the beasts.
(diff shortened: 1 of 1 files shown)


terryyin:
Shall I pull the two pull request together?

rakhimov:
No, they are not meant to be.
Besides, the 'revert-revert' PR should be removed.
I put the comment there.

lizard_languages/clike.py
```diff
@@ -72,16 +72,85 @@ def _typedef(self, _, tokens):
...
+        self._state = self.__declare_structure
+        if token != "if":
+            self._state(token)
+
+    def __declare_structure(self, token):
+        """Ignores structures between parentheses on structure declaration."""
+        if token == "(":
+            self.__paren_count += 1
+        elif token == ")":
+            assert self.__paren_count > 0
```
terryyin: These two `assert`s in production code have caused some trouble for people doing C#.
I've removed them temporarily, but don't know what caused the problem yet. The C# code was from a company so they cannot share it with me.

rakhimov: You don't have to remove them.
C'mon, don't they know how to suppress assertion?
Let them run Python with optimization flag, i.e. -O.
This will suppress asserts for them.

terryyin: But some of them installed lizard using pip, and the users don’t event know it’s Python behind it.

rakhimov: I am a bit confused with the argument.
Installed via pip, and not a Python?
I think the target users are programmers just like we are.
And hopefully, they would appreciate our efforts by reporting bugs.
Lizard is closed-open-source agnostic when comes to analysis.
I bet no company in the world has a proprietary dialect of C# to hide.
The argument that the code is closed and can't report a bug is just phony.

Removing asserts just makes our life harder triaging the bugs.
The users ignoring the asserts/bugs will risk getting invalid results.
It just comes down to fail nosily instead of silently.

terryyin: I agree with your point that errors shouldn’t be ignored silently and it should fail when there’s situation we cannot predict, and hopefully the user will tell us their findings (which happened quite often in the past).

But on the other hand I also hope lizard can be tolerable to broken/half-done code. And by “tolerable” I didn’t mean the error should be swallowed silently but be handled in a proper way.

Let me try to get back to the person who reported this assertion to me and see if I can get the code that triggered the assertion.

Thanks for reminding me this important design principle:-)

rakhimov: ``PYTHONOPTIMIZE`` env variable can also be used.

PR #145, opened 2016-08-08
Update xmloutput.py

JiahangLi:
Try to have the summary of the Avg of CCN on a function base, similar to the txt output format.

### DIFF
--- a/lizard_ext/xmloutput.py
+++ b/lizard_ext/xmloutput.py
@@ -87,8 +87,8 @@ def _create_file_measure(doc, result):
     for key, val in summary:
         measure.appendChild(_create_labeled_value_item(doc, 'sum', key, val))
         
-       summary = [("NCSS", file_total_ncss),
-               ("CCN", file_total_ccn)]
+       summary = [("NCSS", file_total_ncss/file_total_funcs),
+               ("CCN", file_total_ccn/file_total_funcs)]
      for key, val in summary:
         measure.appendChild(_create_labeled_value_item(doc, 'average', key, val))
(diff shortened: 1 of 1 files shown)


JiahangLi:
This build is taking forever. I'm sure there is some issue with it, but I don't know where that occurs.

 What I was trying to do it just change the summary of sum into a summary of average on a function base. (I didn't deleted the sum summary part but mimicked it append the average part to the measure as well.)

I ran my local version and it seems output the right xml results at least but the Jenkins didn't change the text according still.

terryyin:
Thanks:-)

I merged it manually, with a few style improvement.

I know there hasn't been much unit testing for the xmloutput. Is it possible to add some still?

terryyin:
What build? Lizard make or using lizard?

JiahangLi:
Sorry for the late reply. I mean the commit can't pass the CI build thoroughly. It eventually terminated with a job error. 
Big thanks for manually merging it.

I will look into the unit testing later for sure.

I did the try to use CPPNSCC plugin of Jenkins to plot our output.xml, but the table is still printing out the sum CCN and NCSS even after I changed them to average, commented out the sum part as well. The plot is not really giving us information other than function count.(Screenshot attached)
I think it might be some issue of how the plugin interpret with our xml output file. 

plot_ccn
stillsum

Coeur:
@JiahangLi , @terryyin , If this was manually merged, let's close the PR for now. The addition of future unit tests may be tracked as  an "issue" in the meantime if needed.

PR #146, opened 2016-08-15
C++11 function trailing return type

rakhimov:
Trailing return types in C++11 functions
are not registering the function.
The current fix just ignores
whatever is in the trailing return.

---
This change is [Reviewable](https://reviewable.io/reviews/terryyin/lizard/146)


terryyin:
Thanks! I cannot believe that we didn't even have '->' as a token before!

PR #149, opened 2016-08-30
Fix error division by zero when running on empty folder.

milo-hyben:
Hi Terry, 
in case when lizard does not find any function in the file or folder it crashes. 
I added a simple condition to check for such scenarios.

Cheers,
Milo

### DIFF
--- a/lizard_ext/xmloutput.py
+++ b/lizard_ext/xmloutput.py
@@ -87,11 +87,12 @@ def _create_file_measure(doc, result):
     for key, val in summary:
         measure.appendChild(_create_labeled_value_item(doc, 'sum', key, val))
 
-    summary = [("NCSS", file_total_ncss/file_total_funcs),
-               ("CCN", file_total_ccn/file_total_funcs)]
-    for key, val in summary:
-        measure.appendChild(_create_labeled_value_item(
-            doc, 'average', key, val))
+    if file_total_funcs != 0:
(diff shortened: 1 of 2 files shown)


lizard_languages/clike.py
```diff
@@ -135,7 +135,7 @@ def _state_global(self, token):
         if token == "template":
             self._state = self._template_declaration
 
-        elif token in ("struct", "class", "namespace"):
+        elif self.last_token != "." and token in ("struct", "class", "namespace"):
             self._state = self._read_namespace
```
rakhimov: Can elaborate on this?
Some examples and tests?

rakhimov: This seems to be a separate issue from the fixing of division by zero.
I would suggest to put it into another PR.

milo-hyben: Sorry for including that one.
I had reverted back that commit.
It is different issue I will be pushing later, related to Java Class Literals (".class") causing out of memory issue with deep nesting.

terryyin:
@milo-hyben thank you so much for the pull-request. Sorry for the late reply, being travelling for two weeks in a row.

Would you be interested in adding a unit test for this scenario? Or you can tell me one scenario that can reproduce the problem and I will add the test.

milo-hyben:
Hi Terry,
added s simple test for xml output which was causing lizard to crash.
Let me know if you think that is good enough.

terryyin:
Great. Thanks:-)

PR #150, opened 2016-08-30
Add a note of known limitations in code processing

rakhimov:
This note documents some limitations of Lizard
and provides some contract for developers
to keep Lizard's code simpler.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -270,13 +270,25 @@ Limitations
 -----------
 
 Lizard requires syntactically correct code.
-Upon processing input with incorrect or unknown syntax,
-Lizard's behavior is pretty much undefined (in good old C sense).
-It can ignore the code, throw an exception, crash, fail silently, or succeed.
+Upon processing input with incorrect or unknown syntax:
+
+- Lizard guarantees to terminate eventually (i.e., no forever loops, hangs)
+  without hard failures (e.g., exit, crash, exceptions).
+
+- There is a chance of a combination of the following soft failures:
+
+    - omission
+    - misinterpretation
+    - improper analysis / tally
+    - success (the code under consideration is not relevant, e.g., global macros in C)
+
 This approach makes the Lizard implementation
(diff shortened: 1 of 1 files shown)


terryyin:
I think Lizard should aim at "no crashes and no dead loops" for incorrect & incomplete code. But doesn't need to provide warning or error message for incorrect & incomplete code.
And I should try to achieve "nearly linear correlation" to the completeness of the code.
So that potentially lizard can work at the background of an editor in real time.

What do you think?

rakhimov:
That's an interesting goal I didn't consider. Let me think.

rakhimov:
I think I went to far requiring the behavior to be undefined.
In case of real time processing, I think your idea of soft guarantees on best effort is better.
I should also note that "undefined behavior" and Python tool doesn't sound good.

README.rst
```diff
@@ -265,6 +265,29 @@ before a function it will suppress the warning for that function.
...
    }
 
+
+Limitations
+-----------
+
+Lizard requires syntactically correct code.
+Upon processing input with incorrect or unknown syntax,
+Lizard's behavior is pretty much undefined (in good old C sense).
+It can ignore the code, throw an exception, crash, fail silently, or succeed.
```
rakhimov: I am trying to think about how to reword this. The following is just brain dump.
```
In the case of invalid/partial/incorrect/unknown syntax:

- Guarantees (or tries its best to) no hard failures: exit, crash, exceptions.
- There is a chance of a combination of the following soft failures:
    - omission
    - misinterpretation
    - improper analysis / tally
    - success (the code in consideration is not relevant, e.g., global macros in C)

Developers of Lizard attempt to minimize the possibility of soft failures.
Hard failures are bugs in Lizard code,
while soft failures are trade-offs.
```

rakhimov: This means that the behavior is defined.

rakhimov: + (As @terryyin  noted) it guarantees to terminate.

PR #151, opened 2016-09-02
Remove warnings from output if no thresholds exceeded

mdmedley:
The very first time I ran the analyzer I was confused because I saw this big "!!!! Warnings" section with nothing under it. So I removed it. Then I realized that knowing what the thresholds were in your execution is important so I wrote a new print method to give the user that information is a less alarming message. Plus a couple other minor changes to the README.rst and the setup.py files.

### DIFF
--- a/lizard.py
+++ b/lizard.py
@@ -688,6 +688,13 @@ def print_warnings(option, scheme, warnings):
     return warning_count, warning_nloc
 
 
+def print_no_warnings(option):
+    warn_str = "No thresholds exceeded ({0})".format(
+        ' or '.join("{0} > {1}".format(
+            k, val) for k, val in option.thresholds.items()))
+    print("\n" + "=" * len(warn_str) + "\n" + warn_str)
+
+
 def print_total(warning_count, warning_nloc, saved_result, scheme):
     file_infos = list(file_info for file_info in saved_result if file_info)
     all_fun = list(itertools.chain(*(file_info.function_list
@@ -759,8 +766,14 @@ def get_warnings(code_infos, option):
 
 def print_result(result, option, scheme):
     result = print_and_save_modules(result, option.extensions, scheme)
+    warning_count = 0
+    warning_nloc = 0
     warnings = get_warnings(result, option)
-    warning_count, warning_nloc = print_warnings(option, scheme, warnings)
+    if list(warnings):
+        warnings = get_warnings(result, option)
+        warning_count, warning_nloc = print_warnings(option, scheme, warnings)
+    else:
+        print_no_warnings(option)
     print_total(warning_count, warning_nloc, result, scheme)
     for extension in option.extensions:
         if hasattr(extension, 'print_result'):
--- a/README.rst
+++ b/README.rst
@@ -24,8 +24,8 @@ C/C++ header files or Java imports. It can deal with
 -  PHP
 -  Scala
 
-By default lizard will search for any source code that it knows an mix
-all the result together. This might not be what you want. You can use
+By default lizard will search for any source code that it knows and mix
+all the results together. This might not be what you want. You can use
 the "-l" option to select language(s).
 
 It counts
@@ -40,7 +40,7 @@ Functions that exceed these limitations will generate warnings. The exit
 code of lizard will be none-Zero if there are warnings.
 
 This tool actually calculates how complex the code 'looks' rather than
-how complex the code real 'is'. People will need this tool because it's
+how complex the code really 'is'. People will need this tool because it's
 often very hard to get all the included folders and files right when
 they are complicated. But we don't really need that kind of accuracy for
 cyclomatic complexity.
@@ -76,7 +76,7 @@ Usage
 
 ::
 
-   lizard [options] [PATH or FILE] [PATH] ... 
+   lizard [options] [PATH or FILE] [PATH] ...
 
 Run for the code under current folder (recursively):
 
@@ -179,7 +179,7 @@ Analyze a folder recursively: lizard mahjong\_game/src
    ======================================
        66     19    247      1    accept_request@64@./httpd.c
    =================================================================================
-   Total NLOC  Avg.NLOC  Avg CCN  Avg token  Fun Cnt  Warning cnt   Fun Rt   NLOC Rt  
+   Total NLOC  Avg.NLOC  Avg CCN  Avg token  Fun Cnt  Warning cnt   Fun Rt   NLOC Rt
    --------------------------------------------------------------------------------
           554        20     4.07      71.15       27            1      0.04    0.12
 
--- a/setup.py
+++ b/setup.py
@@ -5,7 +5,7 @@ To install lizard:
 '''
 
 import lizard
-from setuptools import setup, find_packages
+from setuptools import setup
 import os
 
 def install(appname):


lizard.py
```diff
@@ -759,8 +766,14 @@ def get_warnings(code_infos, option):
 
 def print_result(result, option, scheme):
     result = print_and_save_modules(result, option.extensions, scheme)
+    warning_count = 0
+    warning_nloc = 0
     warnings = get_warnings(result, option)
-    warning_count, warning_nloc = print_warnings(option, scheme, warnings)
+    if list(warnings):
+        warnings = get_warnings(result, option)
```
rakhimov: This line seems to be a duplicate of line 771.

lizard.py
```diff
@@ -759,8 +766,14 @@ def get_warnings(code_infos, option):
 
 def print_result(result, option, scheme):
     result = print_and_save_modules(result, option.extensions, scheme)
+    warning_count = 0
+    warning_nloc = 0
     warnings = get_warnings(result, option)
-    warning_count, warning_nloc = print_warnings(option, scheme, warnings)
+    if list(warnings):
```
rakhimov: Isn't ``warnings`` always a container of warning (not null)?

```python
if warnings:  # Is there a potential problem?
    ...
```

lizard.py
```diff
@@ -688,6 +688,13 @@ def print_warnings(option, scheme, warnings):
     return warning_count, warning_nloc
 
 
+def print_no_warnings(option):
+    warn_str = "No thresholds exceeded ({0})".format(
+        ' or '.join("{0} > {1}".format(
+            k, val) for k, val in option.thresholds.items()))
```
rakhimov: Code formatting here is a bit confusing.
Why not move the second ``format`` ``k, vall)`` to the same line?

lizard.py
mdmedley: It is a duplicate, but it's because list(warnings) will exhaust the generator that needs to be used by the print_warnings method.

lizard.py
mdmedley: `get_warnings()` returns a generator so `if warnings:` will never be false. So to check that the generator is empty (meaning no warnings) I use `list(warnings)`

lizard.py
mdmedley: This is the exact same formatting as `print_warnings()` on line 679 right above here so I kept it for consistency in hopes that the PR will be more easily adopted.

lizard.py
rakhimov: @mdmedley Thanks for explaining. That was subtle without looking into definition of ``get_warnings`` then ``whitelist_filter``, both of which, sadly, are undocumented.

terryyin:
```
 +    if list(warnings):
 ```
Yes, this line will exhaust the generator. But moving it to get_warnings isn't a good idea since the "warning_only" option still need the result to be a generator.

@rakhimov sorry for the lack of comments for `get_warnings`. It obviously doesn't explain itself very well. What would be a better name?

@mdmedley have you tried adding any unit test for the change? There hasn't been enough UT for the UI. So it will be great if you can add some. Otherwise, I can do it too. Thanks for your contribution!

Comment (rakhimov) [inline lizard.py]: @terryyin some documentation should be enough; otherwise, maybe ``generate_warnings`` would be a better alternative.

Comment (terryyin) [inline lizard.py]: Perhaps get_generator_of_warnings?

Comment (rakhimov) [inline lizard.py]: I am wondering what would be a Pythonic way?

**Comment by mdmedley** [comment]
@rakhimov @terryyin I modified my pull request to remove the duplicate call which prevents the need to exhaust the generator just  recreate it, thus speeding up the application in the cases where there are warnings.

**Comment by mdmedley** [comment]
I also seem to have broken the build. @terryyin or @rakhimov if either of you know exactly where to fix this it would greatly reduce the time it's going to take me to figure it out.

**Comment by rakhimov** [comment]
Testing on Linux kernel code (quad-core i7 laptop):
```bash
lizard -t 6 -Ecpre -a 6 -l cpp > complexity.txt
```
current master branch:
```
real	4m12.183s
user	25m3.672s
sys	0m5.340s
```
Not bad for 12 million lines of code with 21,116 warnings. The output file weighs 48M.

with @mdmedley patch
```
real	 4m9.342s
user 24m45.100s
sys	0m7.672s
```
There seems to be some speedup, maybe?
P.S. It is frying my laptop: 80-86 *C :-)

rakhimov:
The same benchmark as before with pypy:
```bash
>  time pypy ~/projects/lizard/lizard.py -t 6 -Ecpre -a 6 -l cpp > complexity-pypy.txt 

real	1m30.777s
user	9m0.064s
sys	0m7.012s
```
Moreover, it doesn't fry my laptop: 67-70 *C.

lizard.py
```diff
@@ -674,20 +674,31 @@ def clang_warning_format(self):
...
     warning_nloc = 0
     warn_str = "!!!! Warnings ({0}) !!!!".format(
         ' or '.join("{0} > {1}".format(
             k, val) for k, val in option.thresholds.items()))
-    print("\n" + "=" * len(warn_str) + "\n" + warn_str)
-    print(scheme.function_info_head())
-    for warning in warnings:
-        warning_count += 1
+    for warning_count, warning in enumerate(warnings):
+        if warning_count == 0:
```
rakhimov: @mdmedley I think here is the bug. The enumeration is zero-based, so if you have one warning, the warning count is 0. Classical off-by-one.

rakhimov:
As you can notice from the timings, multi-threading is scaling pretty well.

lizard.py
```diff
@@ -674,20 +674,31 @@ def clang_warning_format(self):
...
+            print("\n" + "=" * len(warn_str) + "\n" + warn_str)
+            print(scheme.function_info_head())
         warning_nloc += warning.nloc
         print(scheme.function_info(warning))
+    else:
+        if warning_count is None:
+            print_no_warnings(option)
```
rakhimov: In your else branch, ``warning_count`` is always None, so the check is redundant.

lizard.py
```diff
@@ -674,20 +674,31 @@ def clang_warning_format(self):
...
+            warning_count = 0
     return warning_count, warning_nloc
```
rakhimov: I think you can eliminate the ``warning_count = 0`` assignment in the else branch,
and do it in the return statement. Something like:
```python
return warning_count || warning_count + 1, warning_nloc
```
I am not sure if this works.

lizard.py
```diff
@@ -674,21 +674,22 @@ def clang_warning_format(self):
 
 
 def print_warnings(option, scheme, warnings):
-    warning_count = None
+    warning_count = 0
```
rakhimov: Indeed, this looks better than the original ternary logic.

lizard.py
mdmedley: Thanks @rakhimov this wasn't how I fixed it but you did identify the bug!

lizard.py
```diff
@@ -679,15 +679,26 @@ def print_warnings(option, scheme, warnings):
...
-    for warning in warnings:
+    for _, warning in enumerate(warnings):
+        if warning_count == 0:
+            print("\n" + "=" * len(warn_str) + "\n" + warn_str)
+            print(scheme.function_info_head())
         warning_count += 1
         warning_nloc += warning.nloc
         print(scheme.function_info(warning))
+    else:
+        if warning_count == 0:
```
rakhimov: Can this be non-0?

rakhimov: Looks like gotcha: https://mail.python.org/pipermail/python-ideas/2009-October/006155.html

mdmedley: Yes, line 686 now increments warning_count

rakhimov: Language wart.

rakhimov: If no warnings, it won't enter loop body. Is that true?

mdmedley:
Test ```test_exit_with_non_zero_when_more_warning_than_ignored_number``` is passing on my local branch...
```shell
nosetests test
...............................................................................................................SS................................................................................................................................................................................................................................................................................................................................................................................................
----------------------------------------------------------------------
Ran 497 tests in 0.376s

OK (SKIP=2)```

But I do see that I only have 497 tests while the Travis output has 501 tests.

File: lizard.py (hunk 8 above)
Comment (rakhimov) [inline lizard.py]: My bad, the else is acting like finally.

Comment (mdmedley) [inline lizard.py]: Exactly

**Comment by rakhimov** [comment]
tests are passing. the problem is that for-else wart.

**Comment by mdmedley** [comment]
@rakhimov what do you mean by that I don't understand

File: lizard.py (hunk 8 above)
Comment (rakhimov) [inline lizard.py]: what happens if you just delete ``else``?

Comment (rakhimov) [inline lizard.py]: "... that way may not be obvious at first unless you're Dutch."

Comment (mdmedley) [inline lizard.py]: The tests are still passing locally without the else, so I'm gonna push the change and see if the Travis build will pass.

Comment (mdmedley) [inline lizard.py]: @rakhimov Yay! That worked, thanks!

**Comment by rakhimov** [comment]
There's a good old idiom in C++ to do what ``for-else`` is trying to address: http://llvm.org/docs/CodingStandards.html#turn-predicate-loops-into-predicate-functions

File: lizard.py [hunk 9]
```diff
@@ -679,15 +679,25 @@ def print_warnings(option, scheme, warnings):
     warn_str = "!!!! Warnings ({0}) !!!!".format(
         ' or '.join("{0} > {1}".format(
             k, val) for k, val in option.thresholds.items()))
-    print("\n" + "=" * len(warn_str) + "\n" + warn_str)
-    print(scheme.function_info_head())
-    for warning in warnings:
+    for _, warning in enumerate(warnings):
```
rakhimov: ``enumerate`` isn't needed anymore, is it?

mdmedley: You're absolutely right. Removing...

terryyin:
Thanks for the work and extra clean up:-)

PR #142, opened 2016-08-01
[NOT READY TO MERGE] Python expression structures

rakhimov:
This PR is to provide examples, solicit ideas, find solutions for fixing issue #141.

---

This change is [Reviewable](https://reviewable.io/reviews/terryyin/lizard/142)

PR #155, opened 2016-09-25
Drop the implementation for Python NS metric

### LINKED ISSUE #141: BUG: Python expression structures and nesting level of tokens/structures
Python reutilizes keywords for  list comprehension (`for`, `if`) and ternary operator (`if`, `else`).
the current implementation of the nesting structure count, which depends on the nesting level of tokens,
does not differentiate between expressions or statements for these keywords
and keeps counting the nesting level.

Examples:

``` python
def foo(bar):
  return [x for x in bar if x == 42]

def foo(bar):
  return 42 if bar else -42
```

These examples will produce NS = 2 and 1, respectively, instead of 0.

### LINKED ISSUE #154: Drop implementation for Nested Structure for Python (pylint does it better)
I recently discovered that ``pylint`` has ``max-nested-blocks`` parameter,
and it gets things right. (no issue with #141).
I don't see any point in reinventing the wheel and wasting effort on this feature in Lizard.
Let's drop the NS metric for Python.
This is going to close #141 and #142.

Code-wise, there's nothing needs to change (maybe some warnings, like "Hey, use pylint for that!").
In the readme or theory, this note can be repeated.

rakhimov:
As noted in #154, pylint already provides this metric.
There's no need to duplicate the effort
since pylint is widely available.
There are some non-trivial difficulties implementing NS for Python
with partial parsers as issued in #141.

This commit fixes #141, closes #154, closes #153, closes #142.

---

This change is [Reviewable](https://reviewable.io/reviews/terryyin/lizard/155)


rakhimov:
@terryyin I am not sure where to stick the runtime warning.

rakhimov:
@terryyin Any opinion on this?

terryyin:
please allow me more time. Sorry.

jderehag:
I would hope that you keep it (NS was added to lizard as a thesis work, but with the long-term intention of adding NS metric to https://github.com/jderehag/swat) which heavily relies on lizard (and partial parsers).

The reason for why partial parsers is a requirement in swat is because of how we parse version-control-systems. We essentially do it file-by-file (and version by verison), and for instance in the git case, checking out a complete repo to analyze a single version is simply to costly if you have millions of versions of something.

Apart from that, I think it makes architectural (and end-user) sense to make sure that every metric should be available on all languagues. Otherwise it might end up so that we have very distinctly different parsers (with a potentially different api) for each language, and keeping up with what is supported where is going to be very complicated to explain and maintain.

rakhimov:
>I would hope that you keep it (NS was added to lizard as a thesis work, but with the long-term intention of adding NS metric to https://github.com/jderehag/swat) which heavily relies on lizard (and partial parsers).

@jderehag As the author of the NS implementation in lizard, I am a bit confused with this statement. Are you sure you mean the Nested Structures metric?

jderehag:
@rakhimov oops, I got NS confused with the nesting-depth added by @mehrdad89.
Sorry about that!

rakhimov:
I guess I know how to solve this.
The heuristic would be to consider tokens that don't start with new line an expression
since expressions are preceded by statements,
and Python has one statement per line.
I had missed this major difference that would help separate the token roles.

PR #156, opened 2016-09-28
C++11 Attributes

rakhimov:
Initial support for C++11 [attributes](http://en.cppreference.com/w/cpp/language/attributes).
There's no need to process them for the current Lizard analyses,
so all attributes are ignored.

If attributes are not processed (ignored),
they mess the class/namespace names and confuse the function processing states.

---
This change is [Reviewable](https://reviewable.io/reviews/terryyin/lizard/156)


terryyin:
Excellent!

PR #161, opened 2016-10-18
Refactor setup.py script and the installation

rakhimov:
The setup configurations
and the entry points to the lizard.py are simplified.
System specific bat files are removed.
The `hfcca` entry point is removed.

The changes may break scripts
that use the deleted entry points.

---
This change is [Reviewable](https://reviewable.io/reviews/terryyin/lizard/161)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,7 @@ install:
 install:
   - pip install -e .
   - pip install -r dev_requirements.txt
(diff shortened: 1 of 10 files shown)


terryyin:
OK. This is the end of HFCCA...

rakhimov:
Long live Lizard ... : -)

Issue #164, opened 2016-11-04
Change output format (for Visual Studio, among others)


itavero:
The current output format when running `lizard -C 15 -w somefile.c someotherfile.c` can be automatically parsed by IDEs such as Eclipse.
Unfortunately, Visual Studio is having issues with this output format.
It will allow you to click on the file in the Output window to jump to the file, but the line number is not taken into accounting.

The format for Visual Studio is described on MSDN: https://msdn.microsoft.com/en-us/library/yxkt8b26.aspx
It should be something like:
```
path/to/filename.c(123): warning: SomeFunction has 57 NLOC, 23 CCN, 291 token, 1 PARAM, 108 length
```
Currently the only available format is:
```
path/to/filename.c:123: warning: SomeFunction has 57 NLOC, 23 CCN, 291 token, 1 PARAM, 108 length
```

It would be nice to have flag for this specific formatting option and/or have a command line argument to change the format to a custom format.

rakhimov:
I think the title needs to be clarified to address only 'warning format'
because the default output is the summary.

rakhimov:
[cpplint](https://github.com/theandrewdavis/cpplint) supports 'emacs|eclipse|vs7|junit' warning format. The code and tests are open for reference and borrowing if anyone wants to implement these formats.

terryyin:
It seems the only difference is the parenthesis "()"?

That should be easy to do.

PR #165, opened 2016-11-04
Visual Studio compatible output format

itavero:
Possible solution for issue terryyin/lizard#164. :wink:


terryyin:
Thanks @itavero, let me add some tests to it:-)

PR #182, opened 2017-05-18
Add CSV output transform

hallta:
Pretty much what it says: 

Using "--csv" at the command line (I wasn't able to find a logical one-char short switch), will transform the output into CSV.

The output will be identical in structure to the default output, but with some additions: 
* New tokens for each line
* Less verbosity (will not summarize files, will not highlight those that breached a threshold)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -36,3 +36,4 @@ nosetests.xml
 dist/
 .cache/
 .pypirc
+*.sw[a-z]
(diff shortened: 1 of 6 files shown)


terryyin:
Thanks for the pull request. @hallta do you want add a few tests to the new feature?

hallta:
Definitely. I didn't see any for the xml output so wasn't sure how best to write them. Can you point me to an example I can model the test after?

hallta:
Nevermind :) I've figured out how to test this, and added some new tests.

terryyin:
yeah, the xml output wasn't covered by test, unfortunately. Thanks for having a good start:-)

PR #183, opened 2017-05-22
Using "verbose" flag to toggle CSV header

hallta:
A new toggle to include/exclude the CSV header, when using the --csv output type. Tests and README updated appropriately.


terryyin:
Excellent! Thanks:-)

PR #184, opened 2017-05-22
Add JSON type to output transform selections

hallta:
Addition of a new JSON output type, invoked via "--json" switch. Tests added and README updated.


hallta:
Disregard, I'm pulling this request.

PR #241, opened 2018-10-23
adding an _expect_declaration_name state for Swift

### LINKED ISSUE #240: Incorrect behavior when a variable named `get` is used in Swift
I have the following Swift file, but lizard (master branch at b631dfd3561366539633a7279f755e237541f3bf) believes it is one single giant getter:

```
//  AFNetworkingWrapper.swift
//  Created by Antoine Cœur on 20/10/2016.
//  Copyright © 2016 EF Education. All rights reserved.

/*...*/

/// HTTP method definitions.
///
/// See https://tools.ietf.org/html/rfc7231#section-4.3
public enum HTTPMethod: String {
    case options = "OPTIONS"
    case get     = "GET"
    case head    = "HEAD"
    case post    = "POST"
    case put     = "PUT"
    case patch   = "PATCH"
    case delete  = "DELETE"
    case trace   = "TRACE"
    case connect = "CONNECT"
}

extension AFHTTPSessionManager {
    
    convenience init(baseURL url: URL?, requestEncoding: ParameterSerializer?, responseEncoding: ParameterSerializer? = nil) {
        self.init(baseURL: url)
        /*...*/
    }
    
    func foo(_ requestHeaders: [String:String]) {
        /*...*/
    }
    
    func bar(_ requestHeaders: [String:String]) {
        /*...*/
    }
    /*...*/
}
```

lizard output:
```
$ python lizard.py AFNetworkingWrapper.swift 
================================================
  NLOC    CCN   token  PARAM  length  location  
------------------------------------------------
     106     29    714      0     118 get@29-146@/coeur/AFNetworkingWrapper.swift
1 file analyzed.
==============================================================
NLOC    Avg.NLOC  AvgCCN  Avg.token  function_cnt    file
--------------------------------------------------------------
    115     106.0    29.0      714.0         1     /coeur/AFNetworkingWrapper.swift

=========================================================================================
!!!! Warnings (cyclomatic_complexity > 15 or length > 1000 or parameter_count > 100) !!!!
================================================
  NLOC    CCN   token  PARAM  length  location  
------------------------------------------------
     106     29    714      0     118 get@29-146@/coeur/AFNetworkingWrapper.swift
==========================================================================================
Total nloc   Avg.NLOC  AvgCCN  Avg.token   Fun Cnt  Warning cnt   Fun Rt   nloc Rt
------------------------------------------------------------------------------------------
       115     106.0    29.0      714.0        1            1      1.00    1.00
```

Coeur:
This is an untested attempt to fix #240 by adding a declarative state to ignore variable names like `get` or `set`.

### DIFF
--- a/lizard_languages/swift.py
+++ b/lizard_languages/swift.py
@@ -32,7 +32,7 @@ class SwiftStates(CodeStateMachine):  # pylint: disable=R0903
             self._state = self._function_name
         if token == 'init':
             self._function_name(token)
-        if token in ('get', 'set', 'deinit'):
+        if token in ('get', 'set', 'willSet', 'didSet', 'deinit'):
             self.context.start_new_function(token)
             self._state = self._expect_function_impl
         if token == 'protocol':
@@ -40,7 +40,7 @@ class SwiftStates(CodeStateMachine):  # pylint: disable=R0903
         if token in ('let', 'var', 'case', ','):
             self._state = self._expect_declaration_name
 
-    def _expect_declaration_name(self, token):
+    def _expect_declaration_name(self, _):
         self._state = self._state_global
 
     def _function_name(self, token):
--- a/lizard_languages/swift.py
+++ b/lizard_languages/swift.py
@@ -30,7 +30,7 @@ class SwiftStates(CodeStateMachine):  # pylint: disable=R0903
     def _state_global(self, token):
         if token == 'func':
             self._state = self._function_name
-        if token == 'init':
+        if token in ('init', 'subscript'):
             self._function_name(token)
         if token in ('get', 'set', 'willSet', 'didSet', 'deinit'):
             self.context.start_new_function(token)
--- a/lizard_languages/swift.py
+++ b/lizard_languages/swift.py
@@ -37,6 +37,11 @@ class SwiftStates(CodeStateMachine):  # pylint: disable=R0903
             self._state = self._expect_function_impl
         if token == 'protocol':
             self._state = self._protocol
+        if token in ('let', 'var', 'case', ','):
+            self._state = self._expect_declaration_name
+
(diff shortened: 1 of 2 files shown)


Coeur:
There is a Travis test failing on:

                class Time
                {
                    var minutes: Double
                    {
                        get
                        {
                            return (seconds / 60)
                        }
                        set
                        {
                            self.seconds = (newValue * 60)
                        }
                    }
                }

The reason may be that `'\n'` isn't the correct way to identify a newline token?

Coeur:
If a newline can't be tokenized, an alternative may be to leave the state on `:` and `=`.

Coeur:
OK, I found a nice solution by interpreting `,` as a declarative token.
I've also added support for:

````Swift
// https://stackoverflow.com/a/30593673/1033581
extension Collection {
    /// Returns the element at the specified index iff it is within bounds, otherwise nil.
    subscript (safe index: Index) -> Iterator.Element? {
        return indices.contains(index) ? self[index] : nil
    }
}
````

and:

````Swift
    var foo: Int = -1 {
        willSet {
            print(newValue)
        }
        didSet {
            print(foo)
        }
    }
````

I've been through the list of tokens in https://github.com/google/code-prettify/blob/master/src/lang-swift.js (last commit is from me by the way), so I think we're not missing much now.

Coeur:
The Swift tests all pass. The CI jobs (AppVeyor and Travis) are failing for other reasons.

terryyin:
HI @Coeur thanks very much! Is it possible to add several new tests? For the new identifiers and new state.

Coeur:
@terryyin done. The most interesting test is `test_keyword_declarations`, as it's filled with traps.

terryyin:
Thanks!

PR #242, opened 2018-10-24
Fix pylint warning (no-else-return) and (consider-using-in)

Coeur:


terryyin:
Thank you very much for fixing these problems.

PR #243, opened 2018-10-26
render a unique tagCloud order when values are equal

Coeur:
This sorting improvement will prevent codecloud.html from being arbitrarily modified each time we `make`.


terryyin:
What does this fix do?

Coeur:
@terryyin it sorts the keys, then the values: this way, when two tags have the same value they will be sorted instead of being randomized. This change stabilizes the file `codecloud.html`.

PR #248, opened 2019-01-12
Fix js file bugs

GcsSloop:
I found some bug in my js file，and try to fix that。

1. try fix js file have `<code>...</code>` or `<code/>`

2. try fix js file csv output long_name lose '('


terryyin:
@GcsSloop thanks for the fix. Could please also add at least 3 test cases?

GcsSloop:
I add 2 test cases.
but I'm not good at Python, The test cases not all passed in my PyCharm, it has 2 failed like this.

``` python
Error
Traceback (most recent call last):
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/unittest/case.py", line 331, in run
    testMethod()
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/unittest/loader.py", line 32, in testFailure
    raise exception
ImportError: Failed to import test module: test_languages.testPython
Traceback (most recent call last):
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/unittest/loader.py", line 254, in _find_tests
    module = self._get_module_from_name(name)
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/unittest/loader.py", line 232, in _get_module_from_name
    __import__(name)
  File "/Users/gcssloop/WorkSpace/GitHub/lizard/test/test_languages/testPython.py", line 3, in <module>
    from ..testHelpers import get_python_function_list_with_extension
ValueError: Attempted relative import beyond toplevel package
```

It's cause by `from ..testHelpers import get_python_function_list_with_extension`.

I think It was caused by PyCharm configuration error.

terryyin:
Hi @GcsSloop right now it seems many cases are fail on the CI on your branch. Please allow me a bit more time to look at the details. You may also take a look.

GcsSloop:
I fix some test case, but it still have a failed test in my local test. I don't know how to fix it.

qq20190123-173340 2x

> I use "PyCharm for Mac" and Python2.7

I see the CI Console, it seems some error cause by OS and Runtime environment.

=================

by the way，I change a test value for "testCAndCPP.py"

because I think the real value is 3, not 2.  https://stackoverflow.com/questions/15240922/cyclomatic-complexity-with-compound-conditions-and-short-circuiting

```
class Test_Big(unittest.TestCase):

    def test_trouble(self):
        code = "foo<y () >> 5> r;"
        result = get_cpp_function_list(code)
        self.assertEqual(0, len(result))

    def test_typedef(self):
        code = """
        typedef struct tagAAA
        {
        }AAA;

        int func_a(int size)
        {
            if(ccc && eee)
            {
                return 1;
            }
        }
        """
        result = get_cpp_function_list(code)
        self.assertEqual(1, len(result))
        # this function have 'if' and '&&' so it will be 3
        # link: https://stackoverflow.com/questions/15240922/cyclomatic-complexity-with-compound-conditions-and-short-circuiting
        self.assertEqual(3, result[0].cyclomatic_complexity)
```

terryyin:
Sorry for keeping this request open for so long. Let me merge it first and then try to fix the tests.

terryyin:
Regarding the complexity number, you are right it should be 3. And the result from Lizard is also 3. The reason it has been 2 in the test is because of a "bug in the test".

I used a class variable `condition` to hold the predefined condition identifiers for a programming language. They aren't supposed to change. However, this value is changed in some of the test, and left the unexpected side-effect. I've changed the code to prevent this.

terryyin:
Regarding the CSV utf-8 encoder, it seems it's not working yet. So I will leave that out for now. Please send a pull require once it's done. Thanks!

terryyin:
Had to revert the `code_reader.py` regex changes also before of performance issue. Will fix them when I have time.

PR #255, opened 2019-02-18
Fixed travis CI

### LINKED ISSUE #254: underestimating max depth when statements are found in the if block
Hello,

I see an issue with `max_nested_depth` that I reproduced in the following test case:
```
def test_one_function_nd_with_addl_statement_in_if(self):
        result = get_cpp_with_nestdepth("""
        int fun() {
            if(a) {
                b;
                if (c) {
                    d;
                }
            }
        }""")
        self.assertEqual(2, result[0].max_nesting_depth)
```
`result[0].max_nesting_depth` is 1 instead of 2 so this test should fail if you add it to testNestingDepth.py.

Note that without the `b;` statement, I get `max_nesting_depth` 2 which I is correct but I don't believe `b;` should impact the nesting depth.

elmotec:
- Removed comment test in Makefile command line.
- Relaxed requirement for pylint to be defect-free for now.
- Drop support for 3.3. Travis does not seem to find it.
- Skipped (for now) occasionally failing test_basic_duplicate with Python 3.4 and 3.5 only.

I feel more comfortable having a green CI before I make any change to fix  #254. I hope you find it useful too.


terryyin:
Please allow me sometime. Will review this over the weekend.

terryyin:
look good. Thanks!

PR #257, opened 2019-02-22
Nested depth

elmotec:
That's the last pull request for issue #254. I believe it fixes the max_nested_depth value at the cost of stopping the special handling of `for (;;)`. The other 2 are just fixing the test suite.


lizard_ext/lizardnd.py
```diff
@@ -30,7 +30,7 @@ def __call__(self, tokens, reader, l_depth=0):  # pylint: disable=R0912
         if hasattr(reader, "loops"):
             loops = reader.loops
         else:
-            loops = set(['if', 'else', 'foreach', 'for', 'while', '&&', '||',
+            loops = set(['if', 'foreach', 'for', 'while', '&&', '||',
```
terryyin: Why 'else' needs to be removed?

elmotec: I reasoned `else` did not change the nesting level. 

If I put `else` back the code `int fun(){if(a)xx;else b;}` shows a max level of 2 but I believe it should be 1, same as `int fun(){if(a) xx;}`. The relevant test case is `test_one_function_nd_with_else`.

elmotec:
Ah! Just realized I had a broken test when I pulled your new master. Unfortunately, it's more a workaround than a fix.

We go through the same logic for C and Python so code like `int fun(){if(a){def;}}` is going to create a level of nesting for def in Python (which is ok) but also in C (not ok). Not sure what the right way to fix that?

PR #266, opened 2019-06-29
Fixed Issue 265: ignore `for` label name by swift.

kijuky:
This code's ccn is 1.

```swift
something(for: .something)
```

##### remarks

Because I don't know how to scan arrays efficiently in Python, it's an index-based implementation. I would like to point out.


kijuky:
Thank you for merging!

PR #268, opened 2019-06-29
Fixed Issue 267: guard clause is not counted by ccn.

kijuky:
This code's ccn is 2.

```swift
guard isValid else { return }
```

This code is equal `if isValid { return }`, so lizard shold output 2.


terryyin:
Sorry for the delayed response. Thanks!

kijuky:
Thank you for merging!

PR #273, opened 2019-08-22
Fixed warning threshold inconsistency in HTML output.

corentin-plouet:
As documented, warnings are generated when metrics (e.g. CCN) are
bigger (>) than the threshold. Thus, we should use the same logic
in the HTML output, and use greater (>) instead of gte (>=).

### DIFF
--- a/lizard.py
+++ b/lizard.py
@@ -965,7 +965,7 @@ def infer_printer_from_file_ext(path):
 
 def open_output_file(path):
     try:
-        return open(path, 'w', encoding='utf-8')
+        return open(path, 'w')
     except OSError:
         msg = "Error: failed to open output file '{}'\n.".format(path)
         sys.stderr.write(msg)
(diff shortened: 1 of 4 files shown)


corentin-plouet:
Hmm, this was supposed to be two distinct pull requests, but I think I clicked the wrong button...

The second change is actually a new feature, adding a command-line option to specify an output file. This is useful to integrate Lizard in a tool like CMake, which lacks the ability to redirect standard output when running external commands (because it's cross-platform and supports multiple build generators).

Nice project, by the way :+1: I was able to generate a report very easily (aside from the aforementioned limitation), and the word cloud feature is pretty cool.

corentin-plouet:
Well, the test suite worked on my machine...
Looks like I used some Python 3 stuff that's not available in 2 ( open(..., encoding="utf-8" ) )
I'll have to investigate why the MD5 checks fail.

corentin-plouet:
Looks better now :sunglasses:
The MD5 comparison logic was not the right approach: too fragile and hard to maintain. Now I just compare the first line of the files to ensure it's the right format.

terryyin:
This looks cool. thanks!

PR #291, opened 2020-06-27
separate parameter type from identifier

trouvant:
Including whitespace between tokens in the items of `full_parameters` prevents parameter types in explicitly-typed languages being erroneously appended to the front of parameter names.


terryyin:
Smart solution. I will add a test case for this.

Issue #297, opened 2020-08-05
Wrong to scan go func with anonymous function


cyw3:
The existence of anonymous functions in function testCCN will cause function testCCN to be unrecognized and the cyclomatic complexity will be ignored.

```
func testCCN() {
	defer func() {
		println("")
	}()
	if true {
		println("")
	}
	if true {
		println("")
	}
}
```

503440692:
@In latest version, the problem seem still exists. Hope it can be fixed, thanks.

terryyin:
What should be the proper behaviour? Count, by as a separate function, or adding to the outer function?

PR #298, opened 2020-08-06
bugfix-Wrong to scan go func with anonymous function

cyw3:
for #297


terryyin:
There's no new test. It will be great if you can also add a test for this senario. Otherwise I will added later.

cyw3:
got it.

PR #299, opened 2020-08-12
Fix the failing build

dutchedge:
I saw your last build was failing after a pull request #298 from yalechen-cyw/master.
These changes (my mastrer version 1bdf8e2) should give a working build again (I've added a build.sh to work locally and should be safe to ignore.)
Due to my unfamiliarity with travis I needed some commits to get it to work, but the changes in the end are just a few.
Further this is my first pull request so I'm not sure if the patch files get automatically added or if the reference to the working version is included.
* Build failed mainly because one line was too long
* I needed to fix the pygments to version 2.3.1 to have the build on version 3.4 succeed.
* Not sure how you work with the version. I just added a very small increment.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,5 +10,10 @@ python:
 
 os: linux
 
+install:
+  - pip install -e .
+  - pip install -r dev_requirements.txt
+  - python setup.py build install
+
 script:
-  - build.sh
+  - make
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,9 @@ python:
 
 os: linux
 
+virtualenv:
+  system_site_packages: true
+
 install:
   - pip install -e .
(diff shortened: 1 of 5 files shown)


terryyin:
Hi @dutchedge , thanks for fixing the build and sorry for my previous neglecting. The last pull request I got doesn't have a corresponding unit test, so I'm still waiting for the committer to provide a new unit test case.

Yeah, I can see you did many revert commits:-) I'm not a big fan of keeping the git log clean, and I usually don't mind having many small and even irrelevant commits. But in this case since it's very easy, I will do a 'squash and merge' so it will become just one commit.

The versioning is not by change lizard_ext/version.py. Once I add a new version to CHANGELOG.md and than do `make release` it will take the topmost version and overwrite lizard_ext/version.py.

Thanks!

PR #300, opened 2020-08-13
Including coverage, extensions goto_count, statement_count and extended cvs output

dutchedge:
Hi,

I've got a new request for you. Included in this request are:
- code coverage measurement of the executed unit tests
- added a few extensions: goto counter and (a rather basic) statement counter (including tests)
- extended the cvsoutput method to include the extensions (extended the tests for this)
- Tried to work away as much warnings as possible.

If you have any comments on my code please let me know.

One question: Is there a special reason why the tests use assertEqual instead of the assertRegex as stated by the DeprecationWarning?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,7 +4,8 @@ all: extensive pylint
 extensive: tests pep8
 
 tests:
-	py.test test
+	coverage run --source=lizard.py,lizard_ext,lizard_language -m pytest test
+	coverage report -m
 
 tests3:
 	python3 -m unittest test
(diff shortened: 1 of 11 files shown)


dutchedge:
Hi @terryyin,

The initial attempted had a lot of commits again :-). I squashed them to one for your convenience. If there are any other things I can do to make the merge easier. Please let me know.

Regards,

Patric

terryyin:
Hi Patric,

thanks for the pull request. It looks cool. I'm surprised how simple the statement counter is. Perhaps I need a language filter for extensions.

A tiny comment is that `if token == ";":...` could be simplified as ` if token in [';', 'if',...]`.

I think using `assertEqual` is for historical reason. We need to gradually change them all.

thanks again:-)

Issue #202, opened 2017-11-20
Nesting Analysis


hgn:
Is it possible to add an additional feature to count idents? You desribe it itself in https://github.com/terryyin/lizard/blob/master/theory.rst and is probbaly a nice additional metric for software.

terryyin:
Yes, but it will perhaps take some time. I will have some time in the coming weeks.

hgn:
Great! I am really surprised how rock solid lizard runs, really a great piece of software. Thank you!

PR #301, opened 2020-09-03
* Added suggestion of using the simplification as suggested by Terry …

dutchedge:
See the following adjustments:

- Added suggestion of using the simplification as suggested by Terry Yin (if token == ";":... could be simplified as if token in [';', 'if',...].)
- Fixed problem of nested structures and cvs output (max_nested_structures not defined)
- Corrected Statement count
- Added some test code from the pages of Randy Marques. Below the mail that allows me to use the material as test code.
- On Tue, Sep 1, 2020 at 2:39 PM Randy Marques <Randy.Marques@xs4all.nl> wrote:
> Dear Patric, Thank you for your kind request. Very much appreciated. You are granted use op any part of the Generic Build Support documentation as longs as you mention the source. Good luck with your project. Randy Marques

Next to that I have some questions:
- the deprecated assertRegexpMatches cannot be replaced by assertRegex on python 2.7. Do you see another way?
- Python 2.7 has been "endofservice" since 1 januari 2020, see https://www.python.org/doc/sunset-python-2/. How long do you still want to support it?
- Does the implementation of lizardns.py solves [issue 202](https://github.com/terryyin/lizard/issues/202)

### DIFF
--- a/lizard_ext/extension_base.py
+++ b/lizard_ext/extension_base.py
@@ -1,12 +1,11 @@
-'''Base stuff for extensions'''
+"""Base stuff for extensions"""
 from lizard_languages.code_reader import CodeStateMachine
 
 
 class ExtensionBase(CodeStateMachine):
-    '''Base class for all lizard extensions'''
+    """Base class for all lizard extensions"""
 
-    # pylint: disable=W0221
-    def __call__(self, tokens, reader):
+    def __call__(self, tokens, reader=None):
         self.context = reader.context
         for token in tokens:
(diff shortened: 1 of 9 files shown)


terryyin:
Sorry. I've been away and occupied by other things for quite some time.

The commit looks good to me. (a minor thing is
```
c_family = 'c' in reader.language_names or \
                    'cpp' in reader.language_names
```
can be simply filed by arrays intersection.)

Thanks!

PR #306, opened 2020-12-02
add line_to_ccn={line:ccn}

yql70:
helo~
I want a new field to show the relationship between line and cyclomatic complexity, like this: 
image
So I added a variable `line_to_ccn` in the `FunctionInfo` class,  to record lines of code and add up the ccn in `add_condition()`. 
I also output `line_to_ccn` in csv files.

### DIFF
--- a/lizard.py
+++ b/lizard.py
@@ -285,6 +285,7 @@ class FunctionInfo(Nesting):  # pylint: disable=R0902
         self.fan_in = 0
         self.fan_out = 0
         self.general_fan_out = 0
+        self.line_to_ccn = {}
 
     @property
     def name_in_space(self):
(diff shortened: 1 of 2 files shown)


terryyin:
Hi Kylin,

What an interesting idea! Is there a resource I can learn more about it? Thanks!

Sorry I cannot accept the pull request like this. There are two problems:
1. It doesn’t come with tests
2. It should probably be an extension (see https://github.com/terryyin/lizard/blob/master/lizard_ext/lizardgotocount.py <https://github.com/terryyin/lizard/blob/master/lizard_ext/lizardgotocount.py> for example)

I like the idea so I could do it later as well.

Terry

yql70:
This idea references the html output of [dart-code-metrics](https://github.com/wrike/dart-code-metrics), `cyclomaticComplexityLines` records the relationship between line and ccn. 
It may be an extension. Looking forward to adding this feature. Thanks!

PR #313, opened 2021-02-24
Fix wrong variable name in html template

kmiya:
Hi :wave: 
I noticed that all `Parameter count` columns in generated html files are always `0` as the following image.
This PR fixes the issue.
image

ref: https://github.com/terryyin/lizard/commit/4a5d8275521fc86e878c3f3f8a82e0e972d7aa3e


terryyin:
Thanks!

Issue #292, opened 2020-06-29
Incorrect nloc for Python function declarations spanning multiple lines


trouvant:
For example, analysis of the following function
```py
def func(a,
         b):
    print(a)
    print(b)
```
yields `nloc: 2`. 

```py
def func(a,
         b,
         c):
    print(a)
    print(b)
    print(c)
```
yields `nloc: 3`.

I've tested both Python and C++, but it seems specific to Python.

sthagen:
@trouvant the second example given really has three lines of code, right? Three print statements ...

alyssawangqq:
Encountered the same issue as well. This seems to be a bug in Lizard python. When function declarations take more than 1 line, all metrics(LOC, CCN etc) will be calculated only on the declaration part but not implmentation

PR #319, opened 2021-05-14
Fixing bug when python has multiple line for function declaration due to formatting

alyssawangqq:
Currently there's bug in python Lizard when getting LOC and CCN when func declaration has multiple lines. It will only count the function declaration lines but not the implementation. This is because the original code uses indentation to check function ending while for multi-line declaration, the last line will start with ")" with no indent.

Example:
```
def f1(a,
    b
):
  # function implementation for 10 lines
```

Issue in https://github.com/terryyin/lizard/issues/292


terryyin:
Hi @alyssawangqq thanks for the pull request. Could you please add a test for the scenario?

terryyin:
But the new test is not passing?

alyssawangqq:
@terryyin Sorry just fixed error in test

terryyin:
Thank you very much!

PR #320, opened 2021-05-26
Added duplicate code extension to options and fixed typo in readme.

BjrnJhsn:
I could'nt find the duplicate code detector extension in the options so I added it.


terryyin:
Great if you find the duplicate detection extension useful. I haven't finish implementing it to the level that I'm confident to share with large audience yet... that's why it's not in the readme yet...

BjrnJhsn:
Ah, ok, I see. I haven't tried it in depth, but I found it useful. Let's wait with this then.

PR #344, opened 2022-03-02
Make length a property of FileInfo

lpereira95:
Small simplification of `FunctionInfo`'s `length` in order to not care about it anywhere else during the parsing process.

`int(self.start_line != self.end_line)` only for consistency with equal start and end lines.

### DIFF
--- a/lizard.py
+++ b/lizard.py
@@ -281,7 +281,6 @@ class FunctionInfo(Nesting):  # pylint: disable=R0902
         self.full_parameters = []
         self.filename = filename
         self.top_nesting_level = -1
-        self.length = 0
         self.fan_in = 0
         self.fan_out = 0
         self.general_fan_out = 0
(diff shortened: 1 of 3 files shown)


terryyin:
Smart idea. Thanks. Could you please add at least one test case? Because there is a new condition `int(self.start_line != self.end_line)` in the pull request.

lpereira95:
Probably the best way to go is to replace the condition `int(self.start_line != self.end_line)` by `+ 1`. With this, functions that have same start and end lines will have length 1 (which I would say is consistent, as e.g.`start_line=0` and `end_line=1` leads to `length=2` in the previous implementation).

Yesterday I chose to add the condition because some output tests were failing (as `length` was initialize with 0), but now that I think more deeply about it, I think we should simply correct that tests.

What do you think?

terryyin:
Yes, I think correcting the test expectations is the right thing to do.

PR #346, opened 2022-03-16
Fix incorrect python function end line when function declared with multiple lines

lijingda:
fix issue #292 

In a multi-line function declaration, if a subsequent line is indented much more than the first line, it will result in the wrong end line of the function.

Example:
```
def foo(arg1,
        arg2,
        arg3
):
    if True:
        return
# old: 4 now: 6
```


terryyin:
Hi @lijingda thanks for the contribution. Would it be better if we create a new test case for this situation instead of changing the existing test?

And since there is a OR in the new if, perhaps at least another test case is needed?

PR #349, opened 2022-04-01
Update lizardmodified.py

wjonessmd:
Lizard only does modified complexity for languages with switch/case structures (like Java). This change means modified complexity can also be done with languages that have case/when structures (like Ruby).

### DIFF
--- a/lizard_ext/lizardmodified.py
+++ b/lizard_ext/lizardmodified.py
@@ -5,15 +5,24 @@ where the whole switch/case will be counted as 1.
 '''
 
 
+from tokenize import Token
+
+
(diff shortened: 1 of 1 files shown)


terryyin:
Hi  @wjonessmd , thanks for the pull request. I'm sorry that I cannot accept it as is. If it's ok and you are interested I have the following comments:

1. Given the way you write the code, at least 10 new test cases are required (by CCN). I believe 2 or 3 new test cases should be enough, but you have to write code without so much duplicates.
2. We do need new test cases for new logic
3. Try not to use changing variables
4. If you have to use changing variables, try to limit it's scope
5. Try not to compare boolean with True or False

I'm looking forward to your updates:-)

Issue #304, opened 2020-11-13
Failed do detect go func


yql70:
lizard 1.17.7
If there is an anonymous function with a return value type in the testCCN function, it will cause the testCCN function to be undetected.
Even when the latter function testC has a receiver argument, the testC function is not recognized too.

**code：**
``` 
func testCCN() {
	sort.SliceStable(Cells, func(i, j int) bool {
		return Cells[i].SortValue < Cells[j].SortValue
	})
	if true {
		println("")
	}
	if true {
		println("")
	}
}

func(v Vertex) testC()
{
	if true {
		println("")
	}
} 
```
```
 ~/Documents/install/lizard %python3 lizard.py test.go -l go
================================================
  NLOC    CCN   token  PARAM  length  location  
------------------------------------------------
       3      1     24      0       3 bool@2-4@test.go
1 file analyzed.
```

PR #351, opened 2022-04-14
Fix error function statistic in some scenario

lijingda:
fix 2 bug
1. #304 
2. error function information when function return `interface{}`


terryyin:
Thanks for the pull request. It looks great. Thanks.

The `next_if` looks like a bad abstraction now....

PR #359, opened 2022-09-04
Add "--short" option, print in warnings in red or green

bzgec:
1. Add "--short" option prints only warnings and total summary.

2. Print warnings string:
- red (threshold exceeded) - `!!!! Warnings (cyclomatic_complexity > 15 or length > 1000 or nloc > 1000000 or parameter_count > 100) !!!!`
  (problematic functions are not printed in red)
- green (no threshold exceeded) - `No thresholds exceeded (cyclomatic_complexity > 15 or length > 1000 or nloc > 1000000 or parameter_count > 100)`

### DIFF
--- a/lizard.py
+++ b/lizard.py
@@ -29,6 +29,8 @@ import os
 from fnmatch import fnmatch
 import hashlib
 
+import specialTermPrint
+
 if sys.version[0] == '2':
     from future_builtins import map, filter  # pylint: disable=W0622, F0401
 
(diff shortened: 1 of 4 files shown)


terryyin:
Hi @bzgec 
thanks for the pull request and sorry for the late response.
Do you have test for "--short" option now?

bzgec:
Are tests from [02da78e](https://github.com/terryyin/lizard/pull/359/commits/02da78e070f98c0842769070e557b5a03d08c3e0) enough?

terryyin:
Hi @bzgec thanks for the test. I have some comments on the test and solution. Some of them are preventive unless my assumptions are wrong.

1. I think it's an improvement to split `print_and_save_modules` into `save_modules` and `save_modules`. But `save_modules` seems to be a method that is blocking, rather than continuous yielding. The behaviour of lizard will be very different after this change.
2. Since the change is mainly adding one if, having one new test is ok. However, `assertNotEqual` is something I will try my best to avoid. It's too easy to pass and too hard to fail, especially when you are asserting a very specific thing. I find the project so far has 0 assertNotEqual.

PR #366, opened 2022-12-05
added Erlang support

FilipSzm:
https://www.erlang.org/
Hello we use Lizard in our university project and i wanted to add erlang support and it was easiest to add it directly here :) its probly not best and fastest implementation but i think it should work quite good, if my code doesn't meet your requirements i will be happy to fix it :)


terryyin:
Thanks! this looks great. Sorry for the long delay.

PR #371, opened 2023-03-17
ci: Drop support for python versions before v3.8

jdehaan:
As a matter of facts, I could not make the current codebase work at all under python2. IMHO this is also a waste of time and energy.

Current status of the python maintenance for each version.

https://devguide.python.org/versions/

Reasonably, I propose to support from v3.8 upwards, any older v3 might work... I updated the documentation and the CI configurations to have a more modern setup.

NOTE: I did not update the website generation related things. The changeset deals with the lizard tool alone


.appveyor.yml
```diff
@@ -1,7 +1,9 @@
+image: Visual Studio 2022
```
jdehaan, line 1: Defaults otherwise to 2015 if not set explicitly, which does not have a python 3.10...

terryyin:
I haven't been involved much in Python development for years. Is it time to drop Python 2 support?

stkw0:
Absolutely. Python2 have been EOL since 4 years ago. Most Linux distros don't support python2 packages anymore.

PR #381, opened 2023-10-02
PT-109: Add public function count for c#

cmius:


terryyin:
Hi @cmius , thanks for the pull request. I couldn't find any test in the commits?

Issue #352, opened 2022-04-18
lizard gets confused by C++ thousands seperator


dhylands:
If I have a C++ function which uses the thousands seperator (introduced in C++14), then lizard will go looking for a matching single quote and will consider all lines from the ' in 100'000 until the next ' to be part of the same function.

This is a simple file that reproduces the problem:
```C
#include <stdio.h>

int func() {
    return 100'000;
}

// Line 1
// Line 2
// Line 3
// Line 4
// Line 5

void func2() {
    printf("x = '%d'\n", 1234);
}
```
If I run `lizard --length 10` then it incorrectly reports that func is longer than 10 lines:
```
================================================
  NLOC    CCN   token  PARAM  length  location  
------------------------------------------------
      13      1     18      0      13 func@3-15@./foo.cpp
1 file analyzed.
==============================================================
NLOC    Avg.NLOC  AvgCCN  Avg.token  function_cnt    file
--------------------------------------------------------------
     14      13.0     1.0       18.0         1     ./foo.cpp

=========================================================================================================
!!!! Warnings (cyclomatic_complexity > 15 or length > 10 or nloc > 1000000 or parameter_count > 100) !!!!
================================================
  NLOC    CCN   token  PARAM  length  location  
------------------------------------------------
      13      1     18      0      13 func@3-15@./foo.cpp
==========================================================================================
Total nloc   Avg.NLOC  AvgCCN  Avg.token   Fun Cnt  Warning cnt   Fun Rt   nloc Rt
------------------------------------------------------------------------------------------
        14      13.0     1.0       18.0        1            1      1.00    1.00
```
(it considers func to be 13 lines long).

If I add a comment with a single quote in it at the end of line 2, then everything works as expected:
```C
#include <stdio.h>

int func() {
    return 100'000; // '
}

// Line 1
// Line 2
// Line 3
// Line 4
// Line 5

void func2() {
    printf("x = '%d'\n", 1234);
}
```
```
  NLOC    CCN   token  PARAM  length  location  
------------------------------------------------
       3      1      8      0       3 func@3-5@./foo.cpp
       3      1     12      0       3 func2@13-15@./foo.cpp
1 file analyzed.
==============================================================
NLOC    Avg.NLOC  AvgCCN  Avg.token  function_cnt    file
--------------------------------------------------------------
      7       3.0     1.0       10.0         2     ./foo.cpp

=============================================================================================================
[...]

PR #391, opened 2024-08-24
Fix parsing of C++14  ' separator with 0x and 0b

### LINKED ISSUE #360: lizard gets confused by C++ thousands seperator in hex and binary numbers
This is essentially the same bug as #352 except that it applies to using a binary base or hexadecimal base.

i.e. Replace `100'000` in the #352 example with `0b0000'1100` or `0x1234'5678`

salman-javed-nz:
Fixes #360


salman-javed-nz:
The changes in this PR were written using https://github.com/terryyin/lizard/commit/bb8aa4d200a2935b07e6d2bf1e6072b449bc5ae3 as a reference, which fixes a similar bug (#352).

terryyin:
Thanks for fixing this bug @salman-javed-nz .

PR #395, opened 2024-10-07
Fix Golike state machine misinterpretation of func keyword inside Struct and Interface definitions

### LINKED ISSUE #394: Go state machine misinterprets `func` keyword Inside Struct and Interface definitions
Golang state machine is incorrectly interpreting the func keyword inside struct and interface definitions as the start of new function declarations. But, because in golang you cannot define a function body inside type, this leads to incorrect function counts and potentially inaccurate complexity analysis. 

I suppose this issue arises because state machine does not properly handle the context where the func keyword appears inside structs and interfaces. 

Example for `struct`:
```go
type Geometry struct {
    isEqual func(float64, float64) error
 }

func (g *Geometry) sayGoodbye() { } // function name is None
```

Example for `interface`:
```go
type MyComparator struct{}

type Comparator interface {
    Handle(func(int) string)
}

func (m MyComparator) Handle(f func(int) string) {} // function name is None
```

Lexxxzy:
Fixes #394

### DIFF
--- a/lizard_languages/golike.py
+++ b/lizard_languages/golike.py
@@ -13,10 +13,31 @@ class GoLikeStates(CodeStateMachine):  # pylint: disable=R0903
         if token == self.FUNC_KEYWORD:
             self._state = self._function_name
             self.context.push_new_function('')
+        elif token == 'type':
+            self._state = self._type_definition
         elif token in '{':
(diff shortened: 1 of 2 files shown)


terryyin:
Thanks for fixing the bug @Lexxxzy .

PR #396, opened 2024-10-10
CSV Export: Adapt for multiple column_names in FUNCTION_INFO

w31t1:
Some extensions like IO (Fan-In/Fan-Out) generate more than one data column. All these columns shall appear in the CSV output.

### DIFF
--- a/lizard_ext/csvoutput.py
+++ b/lizard_ext/csvoutput.py
@@ -24,14 +24,11 @@ def csv_output(result, options):
     extension_captions = []
     for extension in options.extensions:
         if extension.__class__.__name__ == 'LizardExtension':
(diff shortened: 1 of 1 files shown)


w31t1:
Tests executed on Python 3.12.5 successfully.

I considered an additional security check in L27 (`if hasattr(...)`):

```
            if hasattr(extension, 'FUNCTION_INFO') and isinstance(extension.FUNCTION_INFO, dict):
```

but removed it as the data structure is defined in class `OutputScheme`.

w31t1:
Unfortunately I cannot identify the build failures raised by the continuous integration pipeline. Can somebody point me to the problem?

Issue #392, opened 2024-08-24
"Cannot find the lizard_languages module."


ccsyt1205:
Dear author, I downloaded the python package and the lizard package, and I tried to run lizard, but I encountered a problem "Cannot find the lizard_languages module". Just as following:
![image](https://github.com/user-attachments/assets/f5f9d53a-f72e-4e02-89f6-fc01ad5c0c82)
How can I solve it?

SshShamma:
I am also getting this error in latest Lizard version.

terryyin:
I've made a new blank release. Please try again. Let's see if it's a release problem.

ccsyt1205:
I retried on my windows，and it still couldn't work. But it works on Linux.

---- Replied Message ----
| From | Terry ***@***.***> |
| Date | 04/23/2025 14:24 |
| To | terryyin/lizard ***@***.***> |
| Cc | ccsyt1205 ***@***.***>,
Author ***@***.***> |
| Subject | Re: [terryyin/lizard] "Cannot find the lizard_languages module." (Issue #392) |

I've made a new blank release. Please try again. Let's see if it's a release problem.

—
Reply to this email directly, view it on GitHub, or unsubscribe.
You are receiving this because you authored the thread.Message ID: ***@***.***>

terryyin left a comment (terryyin/lizard#392)

I've made a new blank release. Please try again. Let's see if it's a release problem.

PR #399, opened 2024-10-31
require pygments

dtrodrigues:
`erlang.py` requires `pygments`, so make it a requirement of the project.

Resolves https://github.com/terryyin/lizard/issues/392


terryyin:
sigh. I guess that will mark the end of lizard with 0 dependencies for production code.

Thanks.

Issue #358, opened 2022-06-29
Is it accurate to get the result data of fan-in fan-out


YesonWyld:
Is it accurate to get the result data of fan-in fan-out, when I used  "lizard -E io". I'm not sure that because the "io" parameter is not mentioned in the description of -E in help message.

Tsh98:
Hello, please tell me how to calculate fan in Python_ in fan_ Out, "lizard. Analyze_file (file_path)" analyze a java file to find the function fan_ in fan_ Out results are all 0

Issue #250, opened 2019-01-22
Fail to calculate fan_in and fan_out


ishepard:
I am having problems in calculating `fan_in` and `fan_out` of the methods. I tried different languages, different approaches (analyze recursively a directory, analyze only one file, etc.), but the result is always the same: `fan_in` and `fan_out` = 0.

What am I doing wrong? 

This is the code I am using:
```
import lizard

an = lizard.analyze_file("/path/to/HTTP.c")
# also tried analyze, analyze_file and giving the directory

print(an.filename)
for func in an.function_list:
    print(func.fan_in)
    print(func.fan_out)
```
the output is always 0.

terryyin:
The fan_in/out feature is an extension of Lizard, and I think at this moment it’s not working.

hgn:
Yes, it is not working yet. I realized that too. ;-) Beside this: thank you @terryyin for lizard and the effort in developing that tool!

Issue #102, opened 2016-03-02
Henry's & Kafura's Fan-In Fan-Out Complexity metric


mehrdad89:
I was wondering if it would be interesting to add this metric as another feature for Lizard. 
In short, Fan-in is the number of links coming into a node, whereas fan-out is the number of arrows going out of a node.
I'll be happy to contribute if you are also interested @terryyin?

Thanks in advance.

terryyin:
Yes, I have an ambition of implementing the Internal Relationship Diagram as recommended in @ michaelfeathers 's book (Working Effectively With Legacy Code).

Having Fan-In and Fan-Out is a good start towards that direction.

mehrdad89:
Great!
I will look into this metric more in a detailed manner, however, I was wondering if you are fine with including an option for the relationship between classes and functions in a project scoop. 
Unfortunately, this may lead to caring about headers and import in source code.
In the meantime, I will continue working on adding new support for nesting depth metric too.

terryyin:
Right, Scoping the biggest problem. So to have fan-in and fan-out is a more pragmatic first step.

I’ve moved a bit of the Nesting Depth code to ext already. Will continue to more and make the ext interface more friendly to both users and us developers.

mehrdad89:
Great! Thanks for that.
I will continue working on both metrics and how to implement them as of today then.

tobias-klein:
I'd be interested in the fan-in fan-out metrics as well! :)

mehrdad89:
I am working on it. 
hopefully, it will be finished next week.

mehrdad89:
@terryyin I was wondering what is your opinion on how print_fan_in_fan_out should be handled.
I just finished writing a working prototype of the fan-in fan-out function in the `lizard.py`. you can check it out in my forked repository [named](https://github.com/mehrdad89/lizard/blob/master/lizard.py) 
I am open to suggestion if you are interested.

mehrdad89:
@terryyin I just used an example from your own repository i.e `mahjong\src\html_ui\html_game.c`
this is the output:
`================================================
  NLOC    CCN   token  PARAM  length  location  
------------------------------------------------
       9      2     56      2      11 start_new_player@26-36@C:\mahjong-master\src\html_ui\html_game.c
       4      1     20      2       4 do_user_do_not_exist_error@38-41@C:\mahjong-master\src\html_ui\html_game.c
[...]

PR #401, opened 2024-11-19
WIP: Implement new Fan-In/Out/General

w31t1:
This PR replace the old Fan-In/Out/General-Out metric by a new approach used in my projects. 

The current code is partially not working (see #250, #358) still Work in Progress as I discovered in #102. I already state that the current calculation depends on the ordering of the input files.

The new calculation is as described in `lizard_ext/lizardio.py`:

```
        Runs the Lizard Extension for calculation of fan-in and fan-out
        metric in the source code.

        Currently supported languages:
            - CLike:
                - Fan-In:          The count of functions this function is called.
                - Fan-Out:         The count of function called in this function.
                - General Fan-Out: The count of functions called in this function.
            - Python
                - Fan-In:          The count of functions this function is called.
                - Fan-Out:         The count of functions called in this function.
                - General Fan-Out: The count of different functions called in this function.

        Metric definition:
            Fan-In:          The grade of units/entities dependent on this code part inside the unit.
            Fan-Out:         The grade of units/entities this code part depends on.
            General Fan-Out: A more general grade of units/entities this code part depends on.

        Limitation:
            - Metric definition depends on language: Functional languages this metric is different
              calculated than object-oriented languages (because of language design).
            - Metric calculation depends on reader extension: Only CodeReader with implemented
              function `external_dependencies` are used to calculate the fan-out metric.
            - Exact metric calculation depends on not-existent knowledge: Programming languages can
              own a lot of features which are not known at calculation time, e.g.:
                - defines/templates (C/C++)
                - compiler/interpreter introduced hidden function calls (SPARC, x86, PowerPC)
                - different language versions (python {2.7/3.11/...}, **C{99/11/..}**
```

The following tasks are not fulfilled:

1. Fix unit tests, add additional validation tests

This PR is a RFC (Request for Comments). The development is still ongoing, tests will be added for C and Python.

@terryyin : Are you interested in this change?
@mehrdad89 : Do you still want to implement your approach?

It is also a question, if it is better to create a new extension file than canibalize this one.


terryyin:
If you like it you should have put some tests on it.

PR #403, opened 2025-01-09
Support for zig language

jostrzol:
Hi, I wanted to use lizard for my master thesis project, but needed also the support for zig language. I added it here, seems to work. If you let me know that there is something wrong, I can try to fix it.

Side note: in 8f3742c28c9fd5d86fe6adfd906c9c91531390e4 I had to fix some pylint/pycodestyle formatting issues or otherwise `build.sh` would fail on style check. I guess that we just have different versions of pylint/pycodestyle and that's why it failed, but I couldn't find the correct version (`dev_requirements.txt` does not contain versions). I can revert these changes if you'd like.

### DIFF
--- a/lizard_languages/__init__.py
+++ b/lizard_languages/__init__.py
@@ -1,5 +1,6 @@
 ''' programming languages of lizard '''
 
+from lizard_languages.zig import ZigReader
 from .clike import CLikeReader
 from .erlang import ErlangReader
 from .java import JavaReader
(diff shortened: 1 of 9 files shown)


terryyin:
@jostrzol thanks for using lizard and the contribution. The code changes looks great. The test coverage is impressive and the solution is simply elegant. Now I also learn some zig from you. Thanks:-)

Also I think the lint fix looks right. We should add version to dev_requirements.txt, I think.

PR #416, opened 2025-06-23
Fix file reading failure on non-UTF-8 Windows for BOM UTF-8 encoded whitelist and .gitignore files

riku311:
On Windows systems where the default encoding is not UTF-8 (e.g., Japanese cp932), reading UTF-8 BOM files may fail if the encoding is not explicitly specified in the open() function.

## Root Cause
The open() function was called without specifying the encoding, which can lead to failures when reading UTF-8 BOM files on Windows systems with non-UTF-8 defaults like cp932.

## Solution
- Replaced direct file reading with `auto_read()` function for `.gitignore` processing
- Applied same fix to `.whitelizard.txt` file reading
- `auto_read()` function automatically detects and handles UTF-8 BOM files, providing consistent file reading across the codebase

## Testing
Verified successful file reading on both Windows and Ubuntu (WSL) environments with UTF-8 files both with and without BOM.

Also verified behavior when the file does not exist.

Updated test mocks to reflect the replacement of open() with auto_read().

Since a mock object was being passed to the function as a file path, added a whitelist attribute to the mock object for compatibility.

---
``` powershell
(venv) PS D:\dev\lizard> python.exe .\lizard.py
Traceback (most recent call last):
  File "D:\dev\lizard\lizard.py", line 1122, in <module>
    main()
    ~~~~^^
  File "D:\dev\lizard\lizard.py", line 1085, in main
    result = analyze(
        options.paths,
    ...<2 lines>...
        options.extensions,
        options.languages)
  File "D:\dev\lizard\lizard.py", line 62, in analyze
    files = get_all_source_files(paths, exclude_pattern, lans)
  File "D:\dev\lizard\lizard.py", line 969, in get_all_source_files
    _load_gitignore()
    ~~~~~~~~~~~~~~~^^
  File "D:\dev\lizard\lizard.py", line 928, in _load_gitignore
    patterns = [line.strip() for line in gitignore_file.readlines()]
                                         ~~~~~~~~~~~~~~~~~~~~~~~~^^
UnicodeDecodeError: 'cp932' codec can't decode byte 0xef in position 0: illegal multibyte sequence
```
Note: This description was translated and refined using ChatGPT.

### DIFF
--- a/lizard.py
+++ b/lizard.py
@@ -623,7 +623,7 @@ def whitelist_filter(warnings, script=None, whitelist=None):
 
     def get_whitelist(whitelist):
         if os.path.isfile(whitelist):
-            return open(whitelist, mode='r').read()
+            return auto_read(whitelist)
         if whitelist != DEFAULT_WHITELIST:
             print("WARNING: !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!")
             print("WARNING: the whitelist \""+whitelist+"\" doesn't exist.")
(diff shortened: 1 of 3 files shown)


terryyin:
I think `auto_read` is not native to Python, right? What dependencies need to be added?

PR #419, opened 2025-07-31
Rust: correctly interpret lifetimes, labels

jostrzol:
Lizard interpreted two `'` signs as a string instead of a lifetime, hence the bug where the function would be omitted.


terryyin:
Looks good to me. Thanks!

PR #425, opened 2025-09-05
Add Structured Text (IEC 61131-3) language support

WLahn:
Added a reader for Structured Text, an IEC 61131-3 programming language used for PLCs.

ST is used a lot in the automation world, but there are only few static analysis tools out there. 
Having lizard supporting ST is super helpful for code quality, complexity checks, and just getting more visibility into bigger PLC projects.

The declaration and implementation of functions is sepparated in different files, 
parameter count is therefore not supported, but is probably not as relevant in ST code anyway.

Also tested support for nesting depth extension, as this  is a main pain point in our projects.


terryyin:
Thanks for the contribution and sorry for the delay. Was parking it in my browser because it's the first time I learned Structured Text and it looks interesting.

Issue #434, opened 2025-10-18
Add support for PL/SQL


Ollec:
Hello @terryyin,

I would like to add support for PL/SQL.

I have ~10 years of experience with both Python and PL/SQL and work extensively with Oracle database codebases. PL/SQL is widely used in enterprise environments, but currently lacks support in Lizard. This would enable complexity analysis for stored procedures, functions, triggers, and packages. 

I am thinking the scope would be something like:
- Procedures and Functions (standalone and in packages)
- Triggers
- Core control structures (IF/ELSIF, CASE, LOOP, WHILE, FOR)
- Standard PL/SQL file extensions (.sql, .pks, .pkb, .pls, .plb, .pck)

This would be my first contribution to a public repository, so I appreciate your patience. I'm committed to following best practices, writing comprehensive tests, and iterating based on feedback.

Are you open to someone taking a swing at it? 
If yes, do you have any specific contribution guidelines you would like me to follow?
Do you have any contribution guidelines or preferences for the development workflow?
Should I create a draft PR early for architectural feedback, or wait until the implementation is more complete?
Any particular PL/SQL edge cases or features you'd consider critical vs. nice-to-have?

Thank you for your time.

PR #436, opened 2025-10-25
#434 Add support for schema-qualified names and triggers in PL/SQL parser

Ollec:
Thank you @terryyin for adding support for PL/SQ. It was very helpful to see how you implemented it. While testing it on a SQL-heavy repo, I identified some possible improvements.

## Overview
Added support for triggers.

The PL/SQL parser incorrectly identifies schema-qualified function/procedure/trigger names. When analyzing code like:
```sql
CREATE FUNCTION SCHEMA.FUNCTION_NAME RETURN NUMBER IS
```
Reports the function name as "SCHEMA" instead of "FUNCTION_NAME".

## Solution
Added support for dot notation in procedure, function, and trigger name parsing. The parser now:
Detects the . token after encountering what appears to be a name
Reads the next token as the actual function/procedure/trigger name
Handles edge cases like triggers with ON schema.table_name clauses

## Testing
Added new test cases covering procedures, functions, and triggers with schema qualifiers
All tests pass
Verified on real Oracle database schema with 8000+ SQL files
PEP8 compliant

Please let me know if there is anything else I can do to improve the quality of this PR and its acceptance.

### DIFF
--- a/lizard_languages/plsql.py
+++ b/lizard_languages/plsql.py
@@ -1,5 +1,35 @@
 """
 Language parser for PL/SQL (Oracle's Procedural Language extension to SQL)
+
+This module implements complexity analysis for PL/SQL code, supporting:
+- Procedures, Functions, and Triggers
+- Package Bodies (not specifications - they only contain signatures)
+- Nested procedures and functions
+- Anonymous blocks with nested functions (blocks themselves aren't counted)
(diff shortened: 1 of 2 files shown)


terryyin:
Thanks a lot for the contribution.

PR #439, opened 2025-10-31
This change fixes an issue where the line count was incorrect when en…

hankhu:
…countering an empty line following a macro definition continuation marker.

Code to reproduce it.

```c
#define  SQ(x) ((x)*(x)) \

bool foo(status *pstat)
{
    return true;
}
```


terryyin:
@hankhu thanks for the fix. I will add the missing new test after merge your pull request.

PR #450, opened 2025-12-16
fix(lizardns): prevent IndexError when structure_piles is empty

dogmatic69:
The nested structures extension (-ENS) crashed with IndexError when processing code with unbalanced braces from the parser's perspective. This commonly occurred with:

- Swift trailing closure syntax (e.g., '} label: {')
- Swift #Preview macros with multiple trailing closures
- Any code producing standalone '}' or '};' patterns

Root cause:
- _state_global() would pop() from structure_piles on '}', potentially emptying the list
- _block_ending() assumed structure_piles was never empty and accessed [-1] unconditionally, causing IndexError

Fix:
- Guard pop() with emptiness check in _state_global()
- Only transition to _block_ending state if structure_piles is non-empty
- Add early return in _block_ending() if structure_piles is empty

Minimal reproduction case: '};' would crash with:
  IndexError: list assignment index out of range at lizard_ext/lizardns.py line 104

Added Swift-specific tests to prevent regression.

### DIFF
--- a/lizard_ext/lizardns.py
+++ b/lizard_ext/lizardns.py
@@ -70,6 +70,8 @@ class LizardExtension(ExtensionBase):  # pylint: disable=R0903
         self.structure_piles = [0]
 
     def pile_up_within_block(self):
+        if not self.structure_piles:
+            return
         self.structure_piles[-1] += 1
         cur_level = sum(self.structure_piles)
         # Is there a path around _state_global?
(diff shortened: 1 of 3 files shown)


dogmatic69:
Hi @terryyin. What is required to get this merged? (besides the conflict...)

dogmatic69:
I see it was fixed in https://github.com/terryyin/lizard/commit/d368ddc....

Issue #457, opened 2026-02-20
Ruby parser hangs indefinitely on Rails controller files with `%i[]` literals and `&.` safe navigation operator


svivekvarma:
# Ruby parser hangs indefinitely on Rails controller files with `%i[]` literals and `&.` safe navigation operator

## Environment

|                |                                      |
| -------------- | ------------------------------------ |
| lizard version | `1.21.0`                             |
| Python version | `3.12.4`                             |
| OS             | Windows 11 (10.0.26200)              |
| Command        | `python -m lizard <file> --csv -t 1` |

## Description

The Lizard Ruby parser hangs indefinitely (verified >120 seconds, never returns) when analyzing a standard Rails controller file. The file is only ~190 lines with no unusual complexity. The hang occurs regardless of the `-t` threading option (`-t 1`, `-t 2`, `-t 4`, and default all hang).

The bug was discovered during batch analysis of a large Rails monorepo where multiple `.rb` controller and model files were timing out consistently.

## Minimal Reproducer

Save the following as `sample_widget_controller.rb` (~175 lines, generic Rails controller with no proprietary code):

```ruby
# frozen_string_literal: true
class WidgetController < ApplicationController
  include SearchActions
  include GroupSearchActions
  include Sortable

  return_true_for :group_only?
  return_false_for :include_all_in_option?
  helper_method :group_only?, :include_all_in_option?
  protected :group_only?, :include_all_in_option?

  default_section :admin
  bounce_unless :can_manage_widgets?

  available_sorts :widget_search

  before_action :adapt_widget_params, only: %i[add check_duplicate update_type_fields on_provider_update]
  before_action :adapt_search_params, only: [:list]

  def list
    @group_id = params[:search][:group_id] if params[:search] && params[:search][:group_id]
    setup_search
    should_display_table = params[:search].present? || params[:sort].present?
    @presenter = Widget::WidgetListPresenter.new(@search, should_display_table)
  end

  def update_groups_for_list
    setup_search
    render json: { groups: render_to_string(partial: 'groups') }
  end

  def add
    @widget = Widget.new(@widget_params)
    @widget.build_amounts(params[:amounts])
    return unless request.post? && @widget.activate

    flash[:success] = I18n.t(:widget_successfully_added)
    redirect_to widget_list_url
  end

  def check_duplicate
    widget = Widget.new(@widget_params)
    widget.build_amounts(params[:amounts])
    if widget.valid?
      render plain: widget.duplicate?.to_s, layout: false
[...]

PR #458, opened 2026-02-20
The Ruby parser hangs exponentially (verified 2^n) when a file contains `<` (class inheritance) followed by many `?`-suffix method calls (`.nil?`, `.blank?`, `.present?`) — a pattern ubiquitous in Rails code.

svivekvarma:
The Ruby parser hangs exponentially (verified 2^n) when a file contains `<` (class inheritance) followed by many `?`-suffix method calls (`.nil?`, `.blank?`, `.present?`) — a pattern ubiquitous in Rails code.

## Root Cause

The generic type lookahead in `code_reader.py` is vulnerable to catastrophic backtracking:

```python
# Before — [^<>]* can consume '?' chars, creating 2^n backtrack paths
r"|\<(?=(?:[^<>]*\?)+[^<>]*\>)(?:[\w\s,.?]|(?:extends))+\>"
```

When `<` has no matching `>` (Ruby inheritance, not generics), the engine exhausts all ways to distribute `n` `?` symbols across `(?:[^<>]*\?)+` groups.

## Fix

Exclude `?` from the character classes inside the lookahead, eliminating ambiguity:

```python
# After — [^<>?]* cannot consume '?', each \? group must take the next one → O(n)
r"|\<(?=(?:[^<>?]*\?)+[^<>?]*\>)(?:[\w\s,.?]|(?:extends))+\>"
```

Matching behavior for actual generics (`<T?>`, `<? super T>`, `<String?, Integer>`) is unchanged.

## Changes

- **`lizard_languages/code_reader.py`**: Two-char fix — `[^<>]` → `[^<>?]` in both character classes of the lookahead
- **`test/test_languages/testRuby.py`**: Regression test using `threading.Thread` with a 5s join timeout (cross-platform, no `SIGALRM`)

Resolves issue #457

### DIFF
--- a/lizard_languages/code_reader.py
+++ b/lizard_languages/code_reader.py
@@ -162,7 +162,7 @@ class CodeReader:
                 r"|\/\/" + _until_end +
                 r"|\#" +
                 r"|:=|::|\*\*" +
-                r"|\<(?=(?:[^<>]*\?)+[^<>]*\>)(?:[\w\s,.?]|(?:extends))+\>" +
+                r"|\<(?=(?:[^<>?]*\?)+[^<>?]*\>)(?:[\w\s,.?]|(?:extends))+\>" +
                 r"|" + r"|".join(re.escape(s) for s in combined_symbols) +
                 r"|\\\n" +
                 r"|\n" +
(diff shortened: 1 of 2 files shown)


terryyin:
Thanks for the pull request.

PR #466, opened 2026-04-14
Fix TS/TSX/JSX: class methods, type alias FPs, JSX separation

cmius:
## Problem

Several bugs in TypeScript, TSX, and JSX parsing produce incorrect metrics:

1. **TSX/JSX class methods are invisible** — `class Widget { render() {} handleClick() {} }` in `.tsx`/`.jsx` files returns **zero** detected functions. This affects every React class component.

2. **Type aliases produce false positives** — `type Handler = (event: Event) => void;` is detected as a function named `Handler` with CCN=1. Every type alias with an arrow signature is a phantom function in the output.

3. **JSX bundled into TSXReader** — `TSXReader.ext = ['tsx', 'jsx']` means `.jsx` files are processed by the TSX reader. When a tool passes `--languages javascript --languages jsx`, the JSX flag activates TSXReader which also matches `.tsx`, causing double-counting.

4. **TSX arrow functions after type annotations missed** — `const fn: Handler = (x) => x + 1` in `.tsx` files doesn't detect `fn` because the type annotation consumer doesn't replay the boundary token.

### `typescript.py` (429 → 604 lines)

Adds 7 features on top of the upstream base, keeping all upstream logic intact:

| # | Feature | Purpose |
|---|---------|---------|
| 1 | Interface skipping | Don't count `interface Foo { bar(): void }` method signatures as functions |
| 2 | Abstract method handling | Skip `abstract doWork(): void` (no body = not a real function) |
| 3 | Private `#method` syntax | Tokenizer regex for ES2022 `#privateMethod()` |
| 4 | Parameter type filtering | `fn(x: string, y: number)` → 2 params, not 4 |
| 5 | Type alias skip | `type Name = { ... }` blocks skipped entirely |
| 6 | Call vs definition | `const x = someFunc(args)` (call) distinguished from `const x = (p) => body` (definition) |
| 7 | `_in_prop_value` tracking | Typed class fields (`count: number;`) no longer break subsequent method detection |

### `tsx.py` (469 → 174 lines)

Complete rewrite. `TSXReader` now extends `TypeScriptReader` and only overrides tokenization — the duplicated state machine is removed entirely. This fixes:

- **Class method detection** — inherits all of TypeScriptStates' class body parsing
- **CCN accuracy** — single condition-counting path instead of two competing ones
- **JSX attribute expressions** — `style={{ color: 'red' }}` handled via sub-tokenizer delegation instead of fragile brace-counting

### `__init__.py` (+2 lines)

Adds `JSXReader` (import + registration). `TSXReader.ext = ['tsx']`, `JSXReader.ext = ['jsx']` — clean separation.

### Before vs after (run on `master` vs this branch)

| Test case | Before (`master`) | After (this PR) | Bug |
|-----------|-------------------|------------------|-----|
| TSX class: `class Widget { render() {} handleClick() {} }` | `[]` (zero functions) | `[('render', 1), ('handleClick', 1)]` | #1 fixed |
| TSX class with typed field: `class C { state: any = {count:0}; increment() {} render() {} }` | `[]` (zero functions) | `[('increment', 1), ('render', 1)]` | #1 fixed |
| Type alias: `type Handler = (e: Event) => void; type Mapper = (i: string) => number; function realFn() {}` | `[('Handler',1), ('Mapper',1), ('realFn',1)]` — 2 false positives | `[('realFn', 1)]` — correct | #2 fixed |
| TSXReader/JSXReader separation | `TSXReader.ext=['tsx','jsx']`, no JSXReader | `TSXReader.ext=['tsx']`, `JSXReader.ext=['jsx']` | #3 fixed |
| TSX hooks: `function Counter() { const increment = () => setCount(c+1); }` | `[('Counter', 1)]` — misses arrow | `[('increment', 1), ('Counter', 1)]` | #4 fixed |

### Test suite results

```
$ python3 -m pytest test -v
1214 passed, 7 skipped in 1.62s
```

**Zero failures, zero regressions** across all 28 supported languages. The 7 skipped tests are pre-existing (3 in JS for complex nested callbacks, 2 in TS for static methods with return-type annotations, 2 in other languages).

The 5 modified test files alone contribute **214 tests** (209 pass + 5 pre-existing skips) covering:
- TS: function declarations, arrow functions, class methods, interfaces, abstract classes, type aliases, generics, typed fields
- TSX: React components (class + functional), JSX attribute expressions, hooks, type annotations, class methods
- JSX: React class components, functional components, CCN in conditionals/switches, arrow functions
- JS: class methods, object methods, getters/setters, async, computed names, ES6 patterns
- ES6: arrow functions, generators, classes, nested functions, computed properties

### Architectural rationale

The key design decision is making `TSXReader` a thin tokenizer layer over `TypeScriptReader` rather than maintaining a parallel state machine. This:

- **Eliminates the class-method bug** — TSX inherits all of TypeScriptStates' class body parsing for free
- **Eliminates CCN double-counting** — single condition path through TypeScriptStates
- **Reduces maintenance** — future TypeScript parser improvements automatically apply to TSX/JSX
- **Reduces code** — 469 → 174 lines, with the 295 removed lines being duplicated logic

## Test plan

- [x] Full upstream test suite: 1214 passed, 7 skipped, 0 failures
- [x] All 28 language parsers tested (C/C++, C#, Go, Java, Kotlin, Lua, ObjC, PHP, Ruby, Rust, Scala, Swift, Perl, Python, Zig, Fortran, Erlang, PLSQL, GDScript, St, R, TTCN, Solidity, TypeScript, TSX, JSX, JavaScript, Vue)
- [x] No modifications to any non-TS/TSX/JSX language parser files
- [x] Before/after comparison confirms each bug is fixed
- [x] Type alias, call-vs-definition, and typed field edge cases covered by dedicated test classes

### DIFF
--- a/lizard_languages/__init__.py
+++ b/lizard_languages/__init__.py
@@ -22,6 +22,7 @@ from .typescript import TypeScriptReader
 from .fortran import FortranReader
 from .solidity import SolidityReader
 from .tsx import TSXReader
+from .tsx import JSXReader
 from .vue import VueReader
 from .perl import PerlReader
 from .st import StReader
@@ -53,6 +54,7 @@ def languages():
         ErlangReader,
         ZigReader,
         TSXReader,
+        JSXReader,
         VueReader,
         PerlReader,
         StReader,
--- a/lizard_languages/tsx.py
+++ b/lizard_languages/tsx.py
@@ -1,27 +1,29 @@
 '''
-Language parser for TSX
+Language parser for TSX/JSX
+
+Uses TypeScriptStates (from typescript.py) for function detection.
+Only overrides tokenization to handle JSX-specific syntax (<Component>, {expressions}).
 '''
 
-from .code_reader import CodeStateMachine
 from .code_reader import CodeReader
 from .js_style_regex_expression import js_style_regex_expression
 from .typescript import TypeScriptReader
-from .typescript import JSTokenizer, Tokenizer, TypeScriptStates
+from .typescript import JSTokenizer, Tokenizer
 
 
 class TSXReader(TypeScriptReader):
     # pylint: disable=R0903
 
-    ext = ['tsx', 'jsx']
-    language_names = ['tsx', 'jsx']
+    ext = ['tsx']
+    language_names = ['tsx']
 
     @staticmethod
     @js_style_regex_expression
     def generate_tokens(source_code, addition='', token_class=None):
-        # Add support for TypeScript type annotations in JSX
         addition = addition + \
             r"|(?:<[A-Za-z][A-Za-z0-9]*(?:\.[A-Za-z][A-Za-z0-9]*)*>)" + \
             r"|(?:<\/[A-Za-z][A-Za-z0-9]*(?:\.[A-Za-z][A-Za-z0-9]*)*>)" + \
+            r"|(?:#\w+)" + \
             r"|(?:\$\w+)" + \
             r"|(?:<\/\w+>)" + \
             r"|(?:=>)" + \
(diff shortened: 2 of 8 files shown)


terryyin:
Why closed?

cmius:
> Why closed?

@terryyin 
This is too big & hard to follow. I have another TWO pr's that i will open soon where i split between bug fixes and enhancements so its easier to follow.

PR #467, opened 2026-04-14
Fix TS/TSX/JSX: missed functions and false positives

cmius:
## Intent

Fix 8 bugs in TypeScript, TSX, and JSX parsing that produce **incorrect metrics output** — missing functions, phantom functions, and inflated complexity scores.

This PR contains **only bug fixes** — no new features, no API changes, no opinionated enhancements. A follow-up PR will add enhancements (interface skipping, abstract methods, `#private` syntax, parameter type filtering).

### 1. TSX/JSX class methods are invisible

**Bug:** `class Widget { render() {} handleClick() {} }` in `.tsx`/`.jsx` files returns **zero** detected functions.

**Root cause:** `JSXTypeScriptStates` duplicated the TypeScript state machine but had no class body parsing. Only standalone `function` keywords and arrow functions were detected.

**Fix:** Rewrite `TSXReader` as a thin tokenizer layer over `TypeScriptReader` (469 → 174 lines), inheriting all class body parsing from `TypeScriptStates`.

| | Before | After |
|--|--------|-------|
| `class Widget { render() {} handleClick() {} }` (.tsx) | `[]` | `[('render', 1), ('handleClick', 1)]` |

### 2. CCN double-counting in TSX/JSX

**Bug:** Both the framework pipeline and `JSXTypeScriptStates` called `add_condition()`, roughly doubling CCN values for `.tsx`/`.jsx` files.

**Fix:** Removing the duplicate state machine (the TSXReader rewrite in bug 1 above) eliminates the double-counting.

### 3. Type alias false positives

**Bug:** `type Handler = (event: Event) => void;` is detected as a function named `Handler` with CCN=1.

**Fix:** Skip `type Name = { ... }` declarations entirely.

| | Before | After |
|--|--------|-------|
| `type Handler = (e) => void; type Mapper = (i) => number; function realFn() {}` | `[Handler, Mapper, realFn]` (2 FPs) | `[realFn]` (correct) |

### 4. Typed class fields break method detection

**Bug:** `class { count: number; method() {} }` detects **zero** methods. The `:` in the field declaration sets `_in_prop_value = True`, but `;`/newline never resets it, so all subsequent `(` tokens are treated as function calls.

**Fix:** Reset `_in_prop_value` on `;`, newline, `,` (object separator), and when a function body pops back to parent scope.

| | Before | After |
|--|--------|-------|
| `class C { state: any = {count:0}; increment() {} render() {} }` (.tsx) | `[]` | `[('increment', 1), ('render', 1)]` |

### 5. Call vs definition confusion

**Bug:** `const x = someFunc(arg1, arg2);` is detected as a function definition.

**Fix:** Compare `last_tokens` to `function_name` and check `_prev_token` — if `(` follows a different identifier than the assigned name, it's a call.

### 6. TSX attribute expression tokenizer broken

**Bug:** `style={{ color: 'red' }}` and nested `{expr}` in JSX attributes cause token stream truncation.

**Fix:** Delegate to `TSXTokenizer` sub-tokenizer for brace-depth tracking instead of manual `brace_count` in `XMLTagWithAttrTokenizer`.

### 7. Async unparenthesized arrow corrupts parser state in class bodies

**Bug:** `field = async x => { ... }` (single parameter without parentheses) causes the parser to lose **all subsequent class methods**. In a 2,700-line Salesforce LWC file with 180+ methods, this drops detection from 184 to 78 named functions.

**Root cause:** The `_async_seen` handler in the class body path captures `=>` before it reaches the arrow function handler, so the arrow function is never registered. The opening `{` is also captured, corrupting brace depth tracking. All methods after the broken arrow are lost.

**Fix:** Exclude `'=>'` from the `_async_seen`/`_static_seen` token capture, and clear both flags in `_arrow_function` so the body's `{` is processed correctly.

| | Before | After |
|--|--------|-------|
| `class C { before() {} handler = async x => { ... }; after() {} }` | `[before]` (after lost) | `[before, handler, after]` |
| Real LWC file (2,743 lines, 184 methods) | 78 named functions | 184 named functions |

### 8. Dot-property field assignment swallows next class method

**Bug:** `field = CONSTANT.PROPERTY;` in a class body causes the immediately following method to be missed. In a Salesforce LWC file with many such assignments, this cascades and loses dozens of methods.

**Root cause:** The `.` token sets `_prev_token = '.'` for chained access detection (`a.b.c()`). But the `;`/newline handler doesn't reset `_prev_token` when it's `'.'` or `'new'` (they're in a preservation exclusion set). So the next method's `(` sees `_prev_token == '.'` and treats it as a method call instead of a definition.

**Fix:** Reset `_prev_token` to `''` on statement boundaries (`;` / newline), unconditionally.

| | Before | After |
|--|--------|-------|
| `class C { x = A.B; m1() {} m2() {} }` | `[m2]` (m1 lost) | `[m1, m2]` |
| Real LWC file (2,957 lines) | 86 named functions | 244 named functions |

### Also fixes
- Static field (`static propTypes = {...}`) swallowing next method
- Abandoned arrow function phantom entries (added `forgive` mechanism)
- `in` vs `==` operator typos for single-char comparisons
- Arrow function double-push (added idempotent guard)
- Method call inside object body missing `sub_state` (caused `)` to escape object reader)
- Docstring incorrectly said "JavaScript" instead of "TypeScript"
- Dead `arrow_function_pending` flag removed (set but never read)

### Known limitation
- Type aliases terminated by newline instead of `;` (e.g., `type ID = string\n`) may swallow the following function. This is not a regression from master — master also misses the function (and additionally produces a false positive). Adding `;` to type aliases avoids the issue.

### Before vs after (run on `master` vs this branch)

| Test case | Before (`master`) | After (this PR) |
|-----------|-------------------|------------------|
| TSX class methods | `[]` (zero functions) | `[('render', 1), ('handleClick', 1)]` |
| TSX class + typed field | `[]` (zero functions) | `[('increment', 1), ('render', 1)]` |
| Type aliases + real fn | `[('Handler',1), ('Mapper',1), ('realFn',1)]` | `[('realFn', 1)]` |
| TSX hooks component | `[('Counter', 1)]` | `[('increment', 1), ('Counter', 1)]` |
| Static field class | `[('render', 1)]` (correct) | `[('render', 1)]` (preserved) |
| Call vs definition | `[('fn', 1)]` (correct) | `[('fn', 1)]` (preserved) |

### Test suite

\`\`\`
$ python3 -m pytest test -v
1211 passed, 10 skipped in 1.23s
\`\`\`

**Zero failures, zero regressions** across all 28 supported languages.

3 of the 10 skips are new tests that document enhancement opportunities (interface skipping, \`#private\` fields) — these will be addressed in a follow-up PR.

## Architectural note

The key change is making `TSXReader` a thin tokenizer layer over `TypeScriptReader` instead of maintaining a parallel `JSXTypeScriptStates` state machine. This:

- Fixes class methods — TSX inherits all of `TypeScriptStates`' class body parsing
- Fixes CCN — single condition-counting path
- Reduces code — 469 → 174 lines
- Reduces maintenance — future TypeScript parser improvements automatically apply to TSX/JSX


terryyin:
Thanks for the contribution. I've rebased the two pull request to trunk so that the merges are not too noisy.

PR #468, opened 2026-04-14
Add TypeScript support for interfaces, #private fields, and field arrow naming

cmius:
## Intent

Add 5 enhancements to the TypeScript/JavaScript parser that improve detection accuracy and report clarity for common TS/JS patterns. These are **additive features**, not bug fixes — the parser works without them, but produces less precise output.

**Depends on:** [PR #467](https://github.com/terryyin/lizard/pull/467) (bug fixes) — this branch builds on top of the bug-fix commit.

### 1. Interface skipping

**Problem:** `interface Service { get(id: string): Promise<Item>; }` produces false-positive function detections — interface method *signatures* are not runtime functions.

**Fix:** Skip entire `interface Foo { ... }` blocks.

| | Without | With |
|--|---------|------|
| `interface Service { get(): Item; create(): Item; } function realFn() {}` | `[get, create, realFn]` (2 FPs) | `[realFn]` |

### 2. Abstract method handling

**Problem:** `abstract doWork(): void;` has no body but is detected as a function.

**Fix:** Track `_in_abstract_context` flag; skip function push when abstract.

| | Without | With |
|--|---------|------|
| `abstract class Base { abstract doWork(): void; concrete() {} }` | `[doWork, concrete]` | `[concrete]` |

### 3. Private `#method` syntax (ES2022)

**Problem:** `#count = 0` confuses the tokenizer, killing subsequent method detection.

**Fix:** Add `#\w+` to tokenizer regex and `#` to identifier validation.

| | Without | With |
|--|---------|------|
| `class C { #count = 0; increment() {} getCount() {} }` | `[]` | `[increment, getCount]` |

### 4. Parameter type filtering with generic support

**Problem:** TypeScript type keywords (`string`, `number`, `boolean`, etc.) pollute `full_parameters` entries. For example, `fn(x: string, y: number)` produces `full_parameters=['x : string', ' y : number']` instead of clean `['x', 'y']`.

Additionally, generic types with commas like `Map<string, number>` cause the comma inside `<...>` to be treated as a parameter separator, producing wrong parameter counts.

**Fix:** Filter `_TS_TYPE_KEYWORDS` from `_dec` and add `<...>` depth tracking to skip commas inside generic type brackets.

| | Without | With |
|--|---------|------|
| `function fn(x: string, y: number, z: boolean) {}` | full_params: `['x : string', ' y : number', ' z : boolean']` | full_params: `['x', 'y', 'z']` |
| `function fn(x: Map<string, number>, y: boolean) {}` | 3 params (generic comma counted) | 2 params (correct) |

### 5. Class field arrow naming

**Problem:** Class field arrows like `handleClick = () => { ... }` are reported as `(anonymous)` even though the field name is available. This makes complexity reports harder to act on — when a threshold is exceeded, the developer has to hunt through the file to find which anonymous function is the culprit.

**Root cause:** The class body `(` handler passed `self.last_tokens` to `_function()`, but after `=`, `last_tokens` has become `'='`, not the field name. The non-class code path (`const handler = () => {}`) already worked correctly because it used `self.function_name`.

**Fix:** When `last_tokens == '='` and `function_name` is set (field assignment pattern), pass `function_name` (the field name) instead of `last_tokens`.

| | Without | With |
|--|---------|------|
| `class Btn { handleClick = () => { ... }; render() {} }` | `[(anonymous), render]` | `[handleClick, render]` |
| `class Form { validate = (v) => { if (!v) return false; }; submit() {} }` | `[(anonymous), submit]` | `[validate, submit]` |

**Why this matters:** Class field arrows are extremely common in modern TS/JS (React class components, event handlers, API clients). Naming them correctly makes complexity reports immediately actionable — `handleClick exceeds CCN 15` is clear; `(anonymous) exceeds CCN 15` requires manual searching.

**What stays anonymous:** Inline callbacks (`.map(x => x * 2)`, `setTimeout(function() {})`) remain `(anonymous)` as expected — they have no assigned name.

**Note on built-in callable filtering:** An earlier version of this PR included a `_JS_BUILTIN_CALLABLES` filter to suppress `const s = String(42)` from being detected as a function definition. This was removed because it was over-broad (suppressed ALL functions named `String`, `Number`, etc., including legitimate definitions and class methods) and because the existing call-vs-definition logic already handles these cases correctly without a separate filter.

## Evidence

```
# All enhancements verified:
interface:        [('realFn', 1, 0)]                       # interface methods skipped
abstract:         [('concrete', 1, 0)]                     # abstract method skipped
private-field:    [('increment', 1, 0), ('getCount', 1, 0)]  # #field handled
param-filter:     fn(x: Map<string, number>, y: boolean) -> 2 params (correct)
field-arrow-name: [('handleClick', 1, 0), ('render', 1, 0)]  # field name, not (anonymous)
builtin-call:     const s = String(42) -> not detected (call-vs-definition handles it)
builtin-def:      function String() {} -> detected correctly (no false negative)
```

## Test suite

```
$ python3 -m unittest test.test_languages.testTypeScript \
    test.test_languages.testES6 test.test_languages.testJavaScript \
    test.test_languages.testJSX test.test_languages.testTSX -v
240 tests, 0 failures, 5 skipped
```

**Zero failures, zero regressions.** Includes 24 new/updated tests across the enhancements:
- 7 tests for class field arrow naming (testTypeScript)
- 4 tests for JS class field arrow naming (testES6)
- 5 tests for builtin-named function detection (testTypeScript)
- 3 tests for abstract method handling (testTypeScript)
- 4 tests for parameter type filtering with generics (testTypeScript)
- 1 updated test for class arrow field (testTypeScript)

## Changes

- `typescript.py`: interface skip block, `_in_abstract_context` flag, `#\w+` tokenizer regex, `_TS_TYPE_KEYWORDS` constant, `_dec` parameter filtering with `<...>` generic depth tracking, field arrow name preservation, removed over-broad `_JS_BUILTIN_CALLABLES` filter
- `tsx.py`: `#\w+` added to tokenizer regex
- `testTypeScript.py`: +227 lines across 4 new test classes
- `testES6.py`: +32 lines, 4 updated/new tests

### DIFF
--- a/lizard_languages/tsx.py
+++ b/lizard_languages/tsx.py
@@ -24,6 +24,7 @@ class TSXReader(TypeScriptReader):
         addition = addition + \
             r"|(?:<[A-Za-z][A-Za-z0-9]*(?:\.[A-Za-z][A-Za-z0-9]*)*>)" + \
             r"|(?:<\/[A-Za-z][A-Za-z0-9]*(?:\.[A-Za-z][A-Za-z0-9]*)*>)" + \
+            r"|(?:#\w+)" + \
             r"|(?:\$\w+)" + \
             r"|(?:<\/\w+>)" + \
             r"|(?:=>)" + \
--- a/lizard_languages/tsx.py
+++ b/lizard_languages/tsx.py
@@ -1,12 +1,14 @@
 '''
-Language parser for TSX
+Language parser for TSX/JSX
+
+Uses TypeScriptStates (from typescript.py) for function detection.
+Only overrides tokenization to handle JSX-specific syntax (<Component>, {expressions}).
 '''
 
-from .code_reader import CodeStateMachine
 from .code_reader import CodeReader
 from .js_style_regex_expression import js_style_regex_expression
 from .typescript import TypeScriptReader
-from .typescript import JSTokenizer, Tokenizer, TypeScriptStates
+from .typescript import JSTokenizer, Tokenizer
 
 
 class TSXReader(TypeScriptReader):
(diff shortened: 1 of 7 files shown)


terryyin:
Thanks for the contribution. I've rebased the two pull request to trunk so that the merges are not too noisy.

PR #472, opened 2026-04-22
fix: prevent IndexError on nested TypeScript template literals

### LINKED ISSUE #471: Crash in duplicate detection on nested TypeScript template literals
When running Lizard 1.22.0 with duplicate detection (-Eduplicate) on valid TypeScript code containing nested template literals, the program crashes with an IndexError: string index out of range. This happens because the tokenizer incorrectly splits nested template strings, yielding an empty string ("") as a token, which breaks the _is_const function in lizardduplicate.py. 

I have developed a quick solution, to prevent the crash, which can be modified or improved later if necessary: in split_template_literal, if brace_count > 0 after scanning the expression (meaning the regex truncated the string prematurely), we simply yield token and return. This returns the chunk as-is instead of generating empty tokens and phantom braces, successfully preventing the crash. I have also written a unit test to verify that the duplicate detector no longer crashes on these constructs. I will open a pull request with this fix shortly!

daruandris:
Fixes #471)

### DIFF
--- a/lizard_languages/typescript.py
+++ b/lizard_languages/typescript.py
@@ -98,6 +98,11 @@ class TypeScriptReader(CodeReader, CCppCommentsMixin):
                     elif content[i] == '}':
                         brace_count -= 1
                     i += 1
+
+                if brace_count > 0:
+                    yield token
+                    return
+
                 expr = content[expr_start:i - 1]
                 yield expr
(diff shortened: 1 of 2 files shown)


terryyin:
Thanks. The fix is great!

PR #474, opened 2026-04-23
fix: avoid double-counting overlapping duplicate ranges

### LINKED ISSUE #473: duplicate_rate double-counts overlapping duplicate ranges and can exceed 100%
I run duplicate detection on official Typescript repo on these directories of src: compiler, services, harness, server, and got 311.99% duplication rate.

The problem is  overlapping duplicate ranges are counted multiple times when duplicate_token_count is aggregated. I fixed this  and got 3.54% which look way better, while the total unique rate remained the same: 97.38%. I will make pull request with the fix shortly.

daruandris:
Fixes: #473. This fixes incorrect duplicate_rate() aggregation when detected duplicate ranges overlap.

Previously, overlapping duplicate ranges were added independently to duplicate_token_count, which could cause the final duplicate rate to exceed 100%.

This patch keeps duplicate detection behavior unchanged, but merges covered duplicate intervals per file boundary before computing the final duplicated token count.

### DIFF
--- a/lizard_ext/lizardduplicate.py
+++ b/lizard_ext/lizardduplicate.py
@@ -2,6 +2,7 @@
 Get Duplicated parameter lists
 '''
 from __future__ import print_function
+from bisect import bisect_right
 from collections import deque
 from itertools import groupby
 from .default_ordered_dict import DefaultOrderedDict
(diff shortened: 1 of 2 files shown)


terryyin:
LGTM, thanks.

PR #477, opened 2026-05-25
Add cyclomatic complexity support for Python `match`/`case` (PEP 634, Python 3.10+)

ixnes:
### Problem

Lizard reported a CCN of 1 for any function containing a `match` statement, regardless of how many `case` arms it had. Each `case` arm is a distinct branch and should count as +1, just like `if`/`elif`.

```python
def classify(cmd):
    match cmd:
        case "quit":   # should be +1
            quit()
        case "go":     # should be +1
            go()
        case _:        # should be +1
            pass
# was: CCN = 1  →  now: CCN = 4
```

### Root cause

`case` and `match` are **soft keywords** in Python — they are only keywords in the context of a `match` statement, and are otherwise valid variable/function names. Because of this, `PythonReader._case_keywords` was left as an empty set, so `condition_counter` never counted `case` tokens.

### Solution

**Soft-keyword detection in `preprocess`** (`lizard_languages/python.py`):

A new `_soft_keyword_lookahead` generator wraps the token stream inside `preprocess`. When it sees `case` or `match` as the first meaningful token on a line, it peeks ahead to decide keyword vs. variable:

| Next token(s) | Decision | Examples |
|---|---|---|
| `=`, `.`, `:`, `,`, `+=` … | variable | `case = 5`, `case.val`, `case: int`, `case, x =` |
| `(` or `[` | scan forward past balanced brackets; keyword if `:` found at depth 0 before end-of-line | `case (x, y):` → keyword; `case(x)` → variable |
| anything else | keyword | `case 1:`, `case _:`, `case {"k": v}:` |

The flags `self._keyword_case` / `self._keyword_match` are set **before** the token is yielded, so both extensions (which run before `process_token`) and `process_token` itself see the correct value for the same token.

**`--modified` extension** (`lizard_ext/lizardmodified.py`):

The existing `--modified` flag makes an entire `switch`/`case` block count as 1. It now treats Python `match`/`case` the same way — adds +1 for the `match` block opener, subtracts -1 per `case` arm, netting +1 for the whole block regardless of arm count. This is symmetric with C/C++/Java `switch`/`case`.

### Disambiguated forms — none of these add complexity

```python
case = 5            # assignment
case.value          # attribute access
foo.case            # member access (not at line start)
case: int = 5       # annotated assignment
case, other = 1, 2  # tuple unpacking
case(x)             # function call
case[0]             # subscript
match = re.match()  # match as variable
match(pattern)      # match as function call
```

### Tests

31 new tests across two classes in `test/test_languages/testPython.py`:

**`Test_Python_match_case`** — regular CCN:
- Basic multi-arm complexity
- `match` alone does not add complexity
- Guards (`case x if cond:`) count the `if` normally
- All variable-use forms correctly produce CCN = 1
- Parenthesized patterns `case (x, y):` and sequence patterns `case [0]:`
- Nested brackets `case (x, (y, z)):`
- Multi-line patterns (implicit line continuation inside brackets)
- Multi-line function calls not counted
- Guard with dict literal in condition (`:` inside `{…}` at depth > 0 must not trigger early)
- Equivalence to `if`/`elif` chain; wildcard arm behaviour

**`Test_Python_match_case_with_modified_ccn`** — `--modified` CCN:
- Entire `match`/`case` block counts as 1 (not 1 per arm)
- Variable uses not penalised under `--modified`
- `match(x)` function call not penalised under `--modified`

### Files changed

| File | Change |
|---|---|
| `lizard_languages/python.py` | `_soft_keyword_lookahead` generator; `_keyword_case` / `_keyword_match` flags; `_SOFT_KW_VARIABLE_NEXT` class attribute |
| `lizard_ext/lizardmodified.py` | `_is_block_keyword` and `_is_case_keyword` helpers for Python soft-keyword awareness |
| `test/test_languages/testPython.py` | 31 new tests |
| `CHANGELOG.md` | Entry under `## Unreleased` |

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,13 @@
 
 ## Unreleased
 
+### New Features
+- **Python** — structural pattern matching (`match`/`case`, PEP 634, Python 3.10+) is now counted correctly:
+  - Each `case` arm adds +1 to cyclomatic complexity (like an `if`/`elif` branch)
+  - `case` guards (`case x if cond:`) count the `if` as a normal condition
+  - `case` and `match` used as plain variable names (assignments, attribute access, function calls, subscripts, tuple unpacking, annotated assignments) are not counted — soft-keyword disambiguation via lookahead in `preprocess`
+  - `--modified`: an entire `match`/`case` block counts as 1, consistent with `switch`/`case` in C-family languages
+
 ## 1.22.2
 
 ### Bug Fixes
--- a/lizard_ext/lizardmodified.py
+++ b/lizard_ext/lizardmodified.py
@@ -7,19 +7,36 @@ where the whole switch/case will be counted as 1.
 
 class LizardExtension(object):  # pylint: disable=R0903
     """
-    Modified CCN extension: counts entire switch/case as 1 complexity.
-    Adds +1 for 'switch', subtracts -1 for each 'case'.
-    Works with switch/case keywords (conceptually from reader.case_keywords).
+    Modified CCN extension: counts an entire switch/case (or Python match/case)
+    as 1 complexity point instead of 1 per arm.
+    Adds +1 for 'switch'/'match' (the block opener), subtracts -1 for each arm.
     """
 
     def __call__(self, tokens, reader):
         for token in tokens:
-            if token == 'switch':  # Add complexity for switch statement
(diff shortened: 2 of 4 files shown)


terryyin:
thanks. It looks great.

PR #480, opened 2026-06-02
fix(objc): handle nested parens in block / function-pointer param types (#365)

### LINKED ISSUE #365: Process error with Objective-C block
```
@implementation TestLizard
+ (UIImage *)sgb_imageWithSize:(CGSize)size drawBlock:(void (^)(CGContextRef context))drawBlock {
    if (!drawBlock) return nil;
    UIGraphicsBeginImageContextWithOptions(size, NO, 0);
    CGContextRef context = UIGraphicsGetCurrentContext();
    if (!context) return nil;
    drawBlock(context);
    UIImage *image = UIGraphicsGetImageFromCurrentImageContext();
    UIGraphicsEndImageContext();
    return image;
}
@end
```
deal with the code above, and get result:
```
Function name -> `sgb_imageWithSize: drawBlock: CGContextRef.image`,
Cyclomatic complexity -> 1
LOC -> 2
```
but in fact, 
```
function name -> `sgb_imageWithSize: drawBlock:`
Cyclomatic complexity -> 3
LOC -> 10
```

https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/ProgrammingWithObjectiveC/WorkingwithBlocks/WorkingwithBlocks.html

StressTestor:
## What

An Objective-C method whose selector takes a **block-typed parameter** is parsed with a garbled name and a wrong line range. Given:

```objc
@implementation TestLizard
+ (UIImage *)sgb_imageWithSize:(CGSize)size drawBlock:(void (^)(CGContextRef context))drawBlock {
    if (!drawBlock) return nil;
    return nil;
}
@end
```

lizard reports the function name as `sgb_imageWithSize: drawBlock: CGContextRef.nil` (the selector with tokens grafted in from the method body) over a collapsed `@10-11`-style range, instead of `sgb_imageWithSize: drawBlock:` over the real method body. Reported as #365.

## Why

`ObjCStates._state_objc_param_type` ends the parameter type at the **first** `)`. A block type `(void (^)(CGContextRef context))` (or a function-pointer type like `(int (*)(int))`) contains nested parentheses, so the scan exits at the inner `)` of `(^)`, desyncs the state machine, and never cleanly enters the method body. Both the selector name and the reported line range fall out of that one desync.

## Fix

Track parenthesis depth in `_state_objc_param_type` so the parameter type ends at its **matching** `)`. Plain parameter types like `(CGSize)size` have no nested parens, so the depth stays 0 and their handling is identical to before.

## Tests

- `test_objc_method_with_block_param` pins both the corrected selector name and the line range (both were corrupted).
- `test_objc_method_with_function_pointer_param` covers the same nested-paren path via a `(int (*)(int))` parameter.

The Objective-C suite passes and the existing plain-selector tests are unaffected.

Fixes #365.

### DIFF
--- a/lizard_languages/objc.py
+++ b/lizard_languages/objc.py
@@ -23,6 +23,10 @@ class ObjCReader(CLikeReader):
 
 
 class ObjCStates(CLikeStates):  # pylint: disable=R0903
+    def __init__(self, context):
+        super(ObjCStates, self).__init__(context)
+        self._objc_param_paren_depth = 0
+
     def _state_global(self, token):
         super(ObjCStates, self)._state_global(token)
         if token == 'typedef':
@@ -50,6 +54,7 @@ class ObjCStates(CLikeStates):  # pylint: disable=R0903
 
     def _state_objc_dec(self, token):
         if token == '(':
+            self._objc_param_paren_depth = 0
             self._state = self._state_objc_param_type
             self.context.add_to_long_function_name(token)
         elif token == ',':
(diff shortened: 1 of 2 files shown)


terryyin:
thanks. The PR look great.

PR #484, opened 2026-06-15
fix(golike): register generic functions with [...] type params

StressTestor:
## what

generic functions/methods with a square-bracket type-param list were dropped or mis-parsed (no name, wrong param count). this registers them correctly.

```scala
def identity[T](x: T): T = { x }   // before: not counted at all
```
```go
func Map[T any](x T) T { return x }   // same
```

## why

in `golike.py`, after a function name `_expect_function_dec` only handled `(` (value params) and `<` (the `<>` generics). a `[` - Scala `def f[T](...)` or Go 1.18 `func F[T any](...)` - fell through to the `else` and bailed back to `_state_global`, so the function was never registered. this affects every golike-based reader: go, scala, kotlin, rust, swift, zig, solidity.

## fix

one branch in `_expect_function_dec` that skips the `[...]` list via `_generalize_type_params`, mirroring the existing `<>` `_generalize`. once the type params are skipped, the following `(` params register the function as normal.

## tests

- `testScala.py`: generic method with one type param, with multiple type params, and a generic method inside a class body (the drop-in-context case)
- `testGo.py`: `[T any]` and multiple type params `[T any, U any]`

## known limitation

this covers the common `def f[T](...)` / `func F[T any](...)` shape. paren-less generic defs like Scala `def empty[T]: List[T]` are still dropped, same as the existing paren-less-def limitation - out of scope here, since closing it means changing how paren-less defs register in general.

### DIFF
--- a/lizard_languages/golike.py
+++ b/lizard_languages/golike.py
@@ -57,11 +57,20 @@ class GoLikeStates(CodeStateMachine):  # pylint: disable=R0903
             self.next(self._function_dec, token)
         elif token == "<":
             self.next(self._generalize, token)
+        elif token == "[":
+            # Square-bracket type parameters, e.g. Scala `def f[T](x: T)` or
+            # Go `func F[T any](x T)`. Skip them like the `<>` generics above
+            # so the following `(` params still register the function.
+            self.next(self._generalize_type_params, token)
         else:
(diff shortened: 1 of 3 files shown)


terryyin:
looks good. sorry for the long delay.

Issue #464, opened 2026-04-09
Feature Request: Halstead Metrics


ArmaanjeetSandhu:
Lizard currently tracks `token_count` but doesn't split tokens into operators vs. operands, so Halstead measures (volume, difficulty, effort, estimated bugs, etc.) can't be derived.

I'd like to add this as a new `-Ehalstead` extension, starting with Python and extending to other languages. The main challenge is that operator/operand classification is language-specific, so I'd want to define a clear interface that language readers can implement.

Is this in scope? And is the `-E` extension mechanism the right place for it, or would you prefer a different approach?

Happy to open a PR if there's interest.

terryyin:
Hi,

Lizard does pass a token stream through to extensions, but those tokens are not classified as operators vs operands, and that classification is not language-neutral. Extensions can use the usual structural context (e.g. current function, nesting), but that alone isn’t enough to compute Halstead metrics reliably.

I think -Ehalstead is the right direction: keep it optional and aligned with other extensions.

For an initial Python-only implementation, it’s fine to keep things simple and self-contained.

For more languages in a coherent way, we may want a small, explicit interface—e.g. a classification hook or a tagged view of the same token stream—implemented per language reader. If that layer stays in one-to-one correspondence with the tokens Lizard already emits, folding it into core later is reasonable. If it needs a different granularity (e.g. not one label per emitted token), I’d prototype that layer inside the extension first until the design stabilizes, rather than baking an immature abstraction into core.

One more thing: Halstead numbers depend on counting rules; whichever convention you follow, a short note in the PR/docs will help users compare results across tools.

Happy to review a PR along these lines.

Terry

PR #485, opened 2026-06-30
feat(halstead): add -Ehalstead extension for Halstead complexity metrics

ArmaanjeetSandhu:
Addresses #464.

## What this adds

An optional `-Ehalstead` extension that computes [Halstead complexity measures](https://en.wikipedia.org/wiki/Halstead_complexity_measures) per function:

```
$ lizard -Ehalstead path/to/code
  NLOC    CCN   token  PARAM  length  H-volume  H-diff  H-effort  location
----------------------------------------------------------------------------
     5      3     29      2       5    116.76    16.9   1973.21 foo@1-5@sample.py
```

Three columns are displayed (`H-volume`, `H-diff`, `H-effort`). The full set (`n1`, `n2`, `N1`, `N2`, vocabulary, length, volume, difficulty, effort, time, estimated bugs) is available programmatically via `function.halstead` and via flat `halstead_*` attributes that also work with `--sort`/`--Threshold` (e.g. `-s halstead_volume`).

## Design

Following the guidance on the issue:

- **`HalsteadClassifier`** is the per-language extension point: `classify(token)` returns operator / operand / skip, one label per token, preserving 1:1 correspondence with the tokens lizard already emits (so this stays easy to fold into core later). A precise `PythonHalsteadClassifier` ships in-tree; a generic C-family classifier is the fallback for other languages.
- Classifier selection (`get_classifier`) checks, in order: a `halstead_classifier` attribute on the reader (the seam intended for per-reader hooks in core), a registry keyed by `language_names`, then the generic fallback.
- **`HalsteadMetrics`** holds the two operator/operand `Counter`s and derives all measures lazily, so values always reflect the final counts.
- Metrics are exposed on `FunctionInfo` as read-only properties, so they default to zero for functions that never went through the extension and integrate with the existing sort/threshold/output machinery.

## Counting convention

Halstead numbers are sensitive to how operators vs operands are counted. In short: operands are identifiers, numeric and string literals, and value-literal keywords (`True`/`False`/`None`, etc.); operators are punctuation/operator symbols plus operator/control keywords (`if`, `for`, `return`, `def`, `and`, …); paired delimiters are counted individually; and tokens are attributed to functions exactly as `token_count` is (so `def`/`class` and the function name belong to the enclosing scope).

## Tests

`test/testHalstead.py` adds 41 tests covering the basic counts, every derived measure (against hand-verified numbers), the flat `FunctionInfo` attributes, operator/operand classification (keywords, literals, strings, numbers, Python value literals and soft keywords), classifier selection precedence, the `HalsteadMetrics` value object, end-to-end C++ via the generic classifier, and extension statelessness. No regressions. Verified working via the console script, `python -m lizard`, and multiprocessing (`-t 2`).

### DIFF
--- a/lizard.py
+++ b/lizard.py
@@ -105,7 +105,8 @@ def _extension_arg(parser):
                         -Ewordcount: count word frequencies and generate tag
                         cloud. -Eoutside: include the global code as one
                         function.  -EIgnoreAssert: to ignore all code in
-                        assert. -ENS: count nested control structures.''',
+                        assert. -ENS: count nested control structures.
+                        -Ehalstead: compute Halstead complexity metrics.''',
                         action="append",
                         dest="extensions",
                         default=[])
--- a/lizard_ext/lizardhalstead.py
+++ b/lizard_ext/lizardhalstead.py
@@ -189,7 +189,8 @@ class HalsteadClassifier(object):  # pylint: disable=too-few-public-methods
         if not token or token.isspace():
             return self.SKIP
         first = token[0]
-        if first in "\"'" or first.isdigit():
+        if first in "\"'" or first.isdigit() or \
+            (first == "." and token[1:2].isdigit()):
             # A string or numeric literal: lizard emits each quoted run or
             # number (or number fragment) as a single token.
             return self.OPERAND
--- a/lizard_ext/lizardhalstead.py
+++ b/lizard_ext/lizardhalstead.py
@@ -1,0 +1,343 @@
+"""
+Halstead complexity measures for lizard (``-Ehalstead``).
+
+Halstead metrics are derived from two basic counts taken over a unit of
+code (here, every function):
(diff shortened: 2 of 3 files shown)


ArmaanjeetSandhu:
@terryyin the CSV issue flagged by the Codex bot is a pre-existing bug, not something Halstead introduced. The same `len == 1` cap was already silently dropping columns for `-Eio` and `-Eduplicated_param_list` as well. I've created a separate PR to fix that.

terryyin:
@ArmaanjeetSandhu got it. Sorry for the delay. Will check and approve one by one.

terryyin:
I also made a small change https://github.com/terryyin/lizard/commit/1ddaeb9a47d0983b8b72a0df7d886f942e0c6b18 to enhance operand classification.

